PR #3, opened 2018-08-10
Update SourceControl.md

chgeuer:
Added links re: cleaning up accidentally published secrets from git history and github repos.


jgardner04:
Related: #5

PR #7, opened 2018-08-20
Added general engineering practices section

max-mironov:
This PR contains list of practices that we try to stick in our day-to-day activities while working on MS [App Center](https://appcenter.ms/) Project.

We believe that this section could also be useful for other teams and this is something that will help everyone to do right things.

Thank you for looking at and let us know your thoughts regarding this.

cc @lukekim @JakubOleksy @keithballinger @matvelloso  @sgrebnov


JakubOleksy:
@keithballinger FYI...something to look at in the context of our 1ES goals.

PR #18, opened 2018-10-07
Added section for Python on secure coding practices

atg-abhishek:
Utilizing a tool like Bandit increases our ability to catch security holes that might be introduced during the development phase early on and mitigate them by following the recommendations to avoid common security pitfalls.

### DIFF
--- a/Engineering/CodeReviews/Python.md
+++ b/Engineering/CodeReviews/Python.md
@@ -24,5 +24,13 @@ We follow the [PEP 257](https://www.python.org/dev/peps/pep-0257/) Docstring con
 
 Flake8 is an option for teams who want a single wrapper for running Pyflakes and pycodestyle.
 
+## Secure Coding
+
+[CSE](../CSE.md) projects should follow secure coding practices adhering to the ethos of "Secure by Design". Security vulnerabilities can be addressed during the development phase by utilizing tools like:
+
+### [```Bandit```](https://github.com/PyCQA/bandit)
+
+Bandit is a tool designed to find common security issues in Python code.
+
 ## Starter Code Review Checklist
 


Engineering/CodeReviews/Python.md
```diff
@@ -24,5 +24,13 @@ We follow the [PEP 257](https://www.python.org/dev/peps/pep-0257/) Docstring con
 
 Flake8 is an option for teams who want a single wrapper for running Pyflakes and pycodestyle.
 
+## Secure Coding
```
seushermsft: When adding a tool, I always worry about conflicts, though I'd hope a security tool and a linter wouldn't conflict. Have you used Bandit and the other recommended tools all using the default configurations? Is there any guidance we should have around configuring Bandit to work without conflicts?

Obviously the other tools mentioned here don't have guidance, but I think they need it, and we need to be prescriptive on what combination of tools has been successful.

atg-abhishek: This is a good point, there isn't a conflict at least from how I've used it in the sense that the linting happens on the fly, continuously in VSCode while Bandit is run after explicitly. That said, I didn't have any specific guidance in mind, the profiles to be used (or you can run all) depend on a prelim analysis of where you think the code that you write might be vulnerable but also an awareness of the profiles that Bandit has to offer. 

That said, as you pointed, given that at the moment we don't actually provide any guidance on the use of the other tools, we might want to identify a template on doing so such that we have consistency in providing a recommended approach for any tool that is shared. Thoughts?

fnocera:
Closing as this has been open for over a year - please feel free to add your contribution to a fresh PR given all the structural changes in the playbook in recent times.

PR #33, opened 2018-12-04
Integrating checks in applications build pipeline

maniSbindra:
This PR add section to show sample docker file for a containerized python application with a stage  to execute the static code analysis (pyflakes, pycodestyle and pydocstyle) against the python code.

### DIFF
--- a/Engineering/CodeReviews/Python.md
+++ b/Engineering/CodeReviews/Python.md
@@ -24,5 +24,63 @@ We follow the [PEP 257](https://www.python.org/dev/peps/pep-0257/) Docstring con
 
 Flake8 is an option for teams who want a single wrapper for running Pyflakes and pycodestyle.
 
+#### Integrating automated checks in the application's build pipeline
+
+For containerized applications one of the ways to include these checks as a part of the application build pipeline is to add a stage for the reviews in the application's Dockerfile as shown below:
+
+```Dockerfile
+# Static code analysis stage / environment
+FROM python:3.7.1-slim AS build-env
+WORKDIR /app
+COPY . ./
+
+RUN pip install pyflakes
+RUN echo "****** Checking source file errors using pyflakes....."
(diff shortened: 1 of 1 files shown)


Engineering/CodeReviews/Python.md
```diff
@@ -24,5 +24,63 @@ We follow the [PEP 257](https://www.python.org/dev/peps/pep-0257/) Docstring con
 
 Flake8 is an option for teams who want a single wrapper for running Pyflakes and pycodestyle.
 
+#### Integrating automated checks in the application's build pipeline
```
seushermsft: I love this!

Another change is proposed which adds a "Secure Coding" heading (#18), and I think to make this useful for that as well, we should create a new heading '## Running Tools' where this example will live, and then when anyone adds a new type of tools, they can update this example.

having great examples like this is what I think CSE needs so we have reusable components that dev crews can just drop into their projects.

Engineering/CodeReviews/Python.md
```diff
@@ -24,5 +24,63 @@ We follow the [PEP 257](https://www.python.org/dev/peps/pep-0257/) Docstring con
...
+# defaults to production, compose may override this to development on build and run
+ARG APP_ENV=production
+ENV APP_ENV $APP_ENV
+
+# default to port 5000 
+ARG PORT=5000
+ENV PORT $PORT
+EXPOSE $PORT
+
+RUN useradd --user-group --create-home --shell /bin/false app &&   mkdir -p /opt/app
```
DanMass: Why three spaces before mkdir?

DanMass:
Please, break this out into a separate file and link to it. Ideally, we could integrate it into something like agogosml ( https://github.com/Microsoft/agogosml ) and point to that example as well.

maniSbindra:
Closing, may reopen new PR, based on new repo structure at a later time

PR #38, opened 2019-02-15
Terraform Template/Starter Project for VS Code

charleszipp:
This is a template for starting a terraform project within vscode. It contains setup for running each of the 5 main terraform commands from the task runner in vscode. We used this with a client to abstract the variable input and for their templates. 

This allows for keeping variable input and backend configuration within the workspace settings; which are not committed to source control. The benefit is that when doing local development devs can keep their settings to themselves safely. Additionally, since the tasks setup all the args for the terraform commands, there's less documentation required for how to run certain commands in their circumstance.

Essentially this is following the npm script pattern to abstract more complex commands + args into shorter aliases.

### DIFF
--- a/Engineering/Templates/Terraform/VSCode/main.tf
+++ b/Engineering/Templates/Terraform/VSCode/main.tf
@@ -1,27 +1,28 @@
-provider "azurerm"{
-    version = "~>1.5"
+provider "azurerm" {
+  version = "~>1.5"
 }
-terraform{
-    backend "azurerm" {
-    }
+
+terraform {
+  backend "azurerm" {}
 }
 
 locals {
-    location_suffixes = {
-        centralus       = "cus"
-        eastus          = "eus"
-        eastus2         = "eus2"
-        westus          = "wus"
-        northcentralus  = "ncus"
-        southcentralus  = "scus"
-        westcentralus   = "wcus"
-        westus2         = "wus2"
-    }
-    location_suffix     = "${local.location_suffixes[var.location]}"
-    suffix              = "-core-${var.env}-${local.location_suffix}-${var.org}"
+  location_suffixes = {
+    centralus = "cus"
+    eastus = "eus"
+    eastus2 = "eus2"
+    westus = "wus"
+    northcentralus = "ncus"
+    southcentralus = "scus"
+    westcentralus = "wcus"
+    westus2 = "wus2"
+  }
+
+  location_suffix = "${local.location_suffixes[var.location]}"
+  suffix = "-core-${var.env}-${local.location_suffix}-${var.org}"
 }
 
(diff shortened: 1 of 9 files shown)


Engineering/Templates/Terraform/VSCode/main.tf
```diff
@@ -0,0 +1,27 @@
...
+    version = "~>1.5"
+}
+terraform{
+    backend "azurerm" {
+    }
+}
+
+locals {
+    location_suffixes = {
+        centralus       = "cus"
```
davidkpiano: Which plugin/extension/??? is used to format the spacing here and make the `=` sign line up? It arguably looks neater, but unless the developer has whatever automatically did this (hopefully not manually), it won't look aligned without manual effort.

My suggestion would be to remove the alignment auto-formatting here for the template. (sorry if this is a bit of a nit)

charleszipp: Hey @davidkpiano, i have updated the .tf files to not have the alignment indicated. To better resolve your point about "which plugin/extension/???", i have added an extensions.json that will cause vs code to recommend the extension favored for terraform development. I believe this should resolve this issue.

DanMass:
When do we expect Fabrikate and Bedrock to subsume all of this? If we can embody some part of the playbook in our tools, I would rather have a pointer to the tools with a quick getting started guide.

charleszipp:
> When do we expect Fabrikate and Bedrock to subsume all of this? If we can embody some part of the playbook in our tools, I would rather have a pointer to the tools with a quick getting started guide.

@DanMass 
My objective with this contribution was to save other crews time in getting up and running with their terraform development in vs code. While on the previous customer project, we didnt land on a solid development pattern for terraform until the last sprint. We struggled with how to support variable inputs on a local machine without having to check into source control the values of those variables. Additionally, with the service principal limitation we found ourselves having to modify templates to support running with actual users from a local box vs. spn's in  azure devops. This was the solution we came up with and thought it would be helpful for other dev crews using tf to deploy any azure resources (not just AKS and k8s). 

While the engineering playbook may not be the best home, I would suggest that the root starter template still have a single source of truth. This will allow it to be evolved as the dependent technologies (terraform and vscode) evolve. Fabrikate and Bedrock could adopt this but, I would advise against having those embody how to establish a terraform project within vscode. This is due to the inevitable divergence that will occur over time as each project moves at its own pace. As a result, the terraform+vscode integration within each project will evolve (in response to changes in terraform and/or vscode) at different rates leading to confusion over which one accurately illustrates the suggested integration pattern. I can move this to a separate github repo if you feel this is not the best home for this.

fnocera:
Please could you update and open a new PR for this given the new playbook structure? Thank you!

PR #43, opened 2019-02-28
Added guidelines for Java code reviews.

polarapfel:
Added guidelines for Java code reviews.

### DIFF
--- a/Engineering/CodeReviews/Java.md
+++ b/Engineering/CodeReviews/Java.md
@@ -1,0 +1,36 @@
+# Java Code Reviews
+
+Please refer to the general [CSE](../../CSE.md) [code review guidance](../CodeReviews.md) as well. Provided here is a check list of Java specific items to help you with your code reviews.
+
+## Code Style
+
+Choose a code style (or coding standard) for your project, stick to it and enforce it. There are several accepted ones to choose from. Good places to start are:
+
+* [Google Java Style Guide](https://checkstyle.org/styleguides/google-java-style-20170228.html)
+* [Oracle's Code Conventions for the Java Programming Language](https://www.oracle.com/technetwork/java/javase/documentation/codeconvtoc-136057.html)
+
+To enforce a coding standard, you can lean on a number of tools for this purpose:
+
+* [checkstyle](http://checkstyle.sourceforge.net/) - "_a development tool to help programmers write Java code that adheres to a coding standard. It automates the process of checking Java code to spare humans of this boring (but important) task._"
+* [PMD](https://pmd.github.io/) - "_a source code analyzer. It finds common programming flaws like unused variables, empty catch blocks, unnecessary object creation, and so forth._"
+
+## Code Quality
+
+Code quality goes hand in hand with code style. Tools like PMD not only uncover style issues, they also point to code quality issues.
+
+You can also use other tools to watch code quality:
+
+* [FindBugs](http://findbugs.sourceforge.net/) - "_a program which uses static analysis to look for bugs in Java code._"
+
+### Code Quality Literature and Guide Lines
+
+There are plenty of good sources to help you with advice and guidance around code quality in Java.
+
+Must read books as a reference for your code reviews include:
+
+* [*Clean Code* by Robert C. Martin](https://www.safaribooksonline.com/library/view/clean-code/9780136083238/)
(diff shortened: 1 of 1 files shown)


Engineering/CodeReviews/Java.md
```diff
@@ -0,0 +1,36 @@
...
+## Code Style
+
+Choose a code style (or coding standard) for your project, stick to it and enforce it. There are several accepted ones to choose from. Good places to start are:
+
+* [Google Java Style Guide](https://checkstyle.org/styleguides/google-java-style-20170228.html)
+* [Oracle's Code Conventions for the Java Programming Language](https://www.oracle.com/technetwork/java/javase/documentation/codeconvtoc-136057.html)
+
+To enforce a coding standard, you can lean on a number of tools for this purpose:
+
+* [checkstyle](http://checkstyle.sourceforge.net/) - "_a development tool to help programmers write Java code that adheres to a coding standard. It automates the process of checking Java code to spare humans of this boring (but important) task._"
```
nmiodice: Another very common static analysis tool is `findbugs` which is apparently now called `spotbugs`. I suggest adding it here.
https://github.com/findbugsproject/findbugs

Engineering/CodeReviews/Java.md
```diff
@@ -0,0 +1,36 @@
...
+* [PMD](https://pmd.github.io/) - "_a source code analyzer. It finds common programming flaws like unused variables, empty catch blocks, unnecessary object creation, and so forth._"
+
+## Code Quality
```
nmiodice: Would it be worthwhile to add guidance on subject areas that are specific to java and/or java-like languages (e.g., C#, Kotlin)? Here are some examples of things that I'm thinking of:

## Dependency Injection
 - Projects of sufficient complexity should consider leveraging dependency injection libraries such as [Spring](https://spring.io/), [Guice](https://github.com/google/guice) or [Dagger](https://google.github.io/dagger/). This typically results in fewer defects by promoting testable code. It also lets you de-couple your configuration and object instantiation from your core business logic.
   - Ex, do you rely on system time? Model this as a dependency using something like [Clock](https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html). Do you rely on an HTTP client? Model this as a dependency using your HTTP client of choice. These dependencies can be injected using your dependency injection framework.

## Interfaces
- If you find yourself creating an interface for which there is only a single implementation (omitting unit tests), you should consider whether or not you need the interface at all. Diligent use of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) can be a good thing.

## Unit Testing
- Pick a standard test framework like [JUnit](https://junit.org/junit5/) or [TestNG](https://testng.org/doc/) and use it as a standard for all builds and integration tests.
- Pick a mocking library like [Mockito](https://site.mockito.org/) and make heavy use of it in integration tests. Mocks should be used to mock dependencies that are injected with your dependency injection framework.
- If you find yourself needing to mock part of the class under test, this is a **strong** indication that you need to refactor your code. Ask yourself whether or not this mocked behavior is better modeled as a dependency.

## Exception Handling
- Are you catching exceptions at the correct layer in your application?
- Is the project using Runtime and Checked exceptions consistently? This is a hotly debated subject, but this is a decision that should be established by the code owners and followed throughout a given codebase. If you are building a library, consider using checked exceptions at the interface level and allowing your clients to decide how to handle them.

nmiodice: Discussed that this content can be added in a later commit PR, but also that some of the content probably lives in a design/testing document and is out of scope for PRs. Given that, I think what we have here is a great base.

mokarian:
I think it's really important to mention the usage of build automation tools. Every java project should benefit form one of these tools to manage project dependencies as well as facilitate project's integration with CI/CD.
tools such as:
Maven: http://maven.apache.org/
Gradle: https://gradle.org/
Ant: http://ant.apache.org/

fnocera:
@polarapfel Thank you so much for this contribution - the playbook has undergone a facelift/structural change, please could you rebase and update the PR. We are working on a process to make approval faster so please reach out once you have updated. Thanks!

fnocera:
Please could you update and open a new PR for this given the new playbook structure? The content should be moved to this location https://github.com/microsoft/code-with-engineering-playbook/tree/master/pull-requests/code-reviews/recipes. Thank you.

PR #71, opened 2019-10-28
Add section about credential scanning

c-w:
This pull request adds a section about credential scanning to the engineering playbook. The section is based on findings from a customer project and answers questions such as:

- What is credential scanning?
- Why do we need credential scanning?
- How can we implement credential scanning?

The section is included as part of the continuous integration heading since for consistency it makes sense to enforce credential scanning policies at the build server level.


fnocera:
@SaraSp @cloudbeatsch flagging for review, thanks!

PR #150, opened 2020-02-15
Feature/111 update python source code guidance

TessFerrandez:
- Updated the python recipe per #111 
- Modified suggestions around linters, reducing the choice to flake8 and pylint as others are already included in flake8
- Expanded on code formatters
- Added section on build validation including example of build validation for python repos
- Added section on pre-commit hooks
- Removed section on Testing as guidance on preferred frameworks is already available in the testing section and doesn't fit here.


pull-requests/code-reviews/recipes/Python.md
```diff
@@ -1,53 +1,207 @@
...
+At the next attempted commit any lint failures will block the commit.
+
+> Note: Installing pre-commit hooks is volontary and done by each developer individually. Thus it is not a replacement for build validation on the server
+
+## Code Review Checklist
+
+In addition to the [Code Review Checklist](../readme.md) you should also look for these python specific code review items
+
+* [ ] Are all new packages used included in requirements.txt
+* [ ] Do funcitons use type hints, and are there any type hint errors?
```
debugerr: nit: typeo functions

pull-requests/code-reviews/recipes/Python.md
```diff
@@ -1,53 +1,207 @@
...
+[CSE](../../../CSE.md) developers follow the [PEP8 style guide](https://pep8.org/) with [type hints](https://www.python.org/dev/peps/pep-0484/). The use of type hints throughout paired with linting and type hint checking avoids common errors that are tricky to debug.
+
+[CSE](../../../CSE.md) projects should check Python code with automated tools.
+
+Linting should be added to build validation, and both linting and code formatting can be added to your pre-commit hooks and VS Code.
 
 ## Linters
 
-[CSE](../CSE.md) projects should check Python code with automated tools. ```pylint``` is the minimum, but we prefer to run three different tools:
+The 2 most popular python linters are [Pylint](https://www.pylint.org/) and [Flake8](https://pypi.org/project/flake8/).  Both check adherance to `PEP8` but vary a bit in what other rules they check.  In general `Pylint` tends to be a bit more stringent and give a few more false positives but both are good options for linting python code.
```
debugerr: nit: I think there are 2 spaces before "Both"

pull-requests/code-reviews/recipes/Python.md
```diff
@@ -1,53 +1,207 @@
...
 
-### [```Black```](https://black.readthedocs.io/en/stable/)
+Both `Pylint` and `Flake8` can be configured in VS Code using the  VS Code `python extension`
```
debugerr: nit: double spaces before "VS Code"

pull-requests/code-reviews/recipes/Python.md
```diff
@@ -1,53 +1,207 @@
...
+def add(first_value: int, second_value: int) -> int:
+    return first_value + second_value
+```
+
+## Build validation
+
+To automate linting with `flake8` and testing with `pytest` in Azure Devops you can add the following snippet to you `azure-pipelines.yaml` file.
+
+```yaml
+trigger:
```
debugerr: I'm back and forth on how much we should include for YAML pipelines. Should we just focus on the actual linter tasks or does it make sense to include triggers here. More content can be easier to setup for ppl, but it's also more static content we need to maintain.

pull-requests/code-reviews/recipes/Python.md
TessFerrandez: Fixed

pull-requests/code-reviews/recipes/Python.md
TessFerrandez: Fixed

pull-requests/code-reviews/recipes/Python.md
TessFerrandez: I'll leave as is for now and we can have a discussion around this in the team checkpoint

code-reviews/recipes/Bash.md
```diff
@@ -4,7 +4,7 @@
 
 [CSE](../../CSE.md) developers follow [Google's Bash Style Guide](https://google.github.io/styleguide/shell.xml).
 
-## Linter
+## Linting
 
 [CSE](../../CSE.md) projects must check Bash code with [shellcheck](https://github.com/koalaman/shellcheck) as part of the [CI process](../../continuous-integration/readme.md).
```
MS-Santi: I think that the Linting sections should be consistent. the other languages state which linters are good and if they have specific installation instructions, how to install.
In this version, the document states it must be part of CI, which seems to be off topic.

On the other hand, I checked the CI document and it does not mention Linting... maybe it should be added there.

code-reviews/recipes/Python.md
```diff
@@ -1,53 +1,208 @@
...
+## Style Guide
 
-[CSE](../CSE.md) developers follow [Google's Python Style Guide](https://google.github.io/styleguide/pyguide.html). Note the use of strong typing throughout. Older versions of python allow you to get away without this but type checking avoids common errors that are tricky to debug. 
+[CSE](../../CSE.md) developers follow the [PEP8 style guide](https://pep8.org/) with [type hints](https://www.python.org/dev/peps/pep-0484/). The use of type hints throughout paired with linting and type hint checking avoids common errors that are tricky to debug.
 
-## Linters
+[CSE](../../CSE.md) projects should check Python code with automated tools.
 
-[CSE](../CSE.md) projects should check Python code with automated tools. ```pylint``` is the minimum, but we prefer to run three different tools:
+Linting should be added to build validation, and both linting and code formatting can be added to your pre-commit hooks and VS Code.
```
MS-Santi: Same as my comment on Bash - if linting should be part of CI, it should be added tot he CI document.

code-reviews/recipes/Python.md
```diff
@@ -1,53 +1,208 @@
...
+The [`Pyright extension`](https://marketplace.visualstudio.com/items?itemName=ms-pyright.pyright) augments VS Code with static type checking when you use type hints
+
+```python
+def add(first_value: int, second_value: int) -> int:
+    return first_value + second_value
+```
+
+## Build validation
+
+To automate linting with `flake8` and testing with `pytest` in Azure Devops you can add the following snippet to you `azure-pipelines.yaml` file.
```
MS-Santi: Is this the right location for this? What if the customer is using Github?

code-reviews/recipes/Bash.md
TessFerrandez: We just had a discussion in another PR on if we should call this section Linting or Linters - and the consensus was Linting so this is just a change to conform to that

code-reviews/recipes/Python.md
TessFerrandez: I think it is good to have a section on this here, as it is tightly coupled with code reviews.
If there is github specific guidance we can add that as well... I will check if I can find some, unless you know of some

code-reviews/recipes/Python.md
TessFerrandez: If it is not already part of the CI/CD documentation and not reported as an issue I think it would probably be good to add it.  Will you? or should I?

code-reviews/recipes/Python.md
TessFerrandez: Update: Added GitHub guidance as well with GitHub Actions -- the YAML is very similar

PR #148, opened 2020-02-13
Updating Readme in Code Reviews

saroup:
With this change I made modification to the readme file in the code-review section and udated the pull request template 
- I updated section about the goals of CRs
- Added more details on healthy code review process.
- refactored some of the other sections.

TODO: There are some sections that still need further details like how to chose PR size, and how to inluce the right SMEs.

#102 #106 #104

PR #153, opened 2020-02-17
Standardize code review templates 

TessFerrandez:
- Updated all recipe docs (except python) to conform to a common standard - Python is changed in #150 
- Update all links for CSE and other links to work after change in PR #148 
- Remove TypeScript recipe as we already have a JavaScript/TypeScript recipe and moved the relevant parts into the JavaScript recipe
- Add link to the Terraform.md from the readme

TODO: this is not verifying or updating technical content - this will be updated in a series of other issues. This is to move the existing content to a consistent format


code-reviews/recipes/CSharp.md
```diff
@@ -1,22 +1,20 @@
 # C# Code Reviews
```
saroup: ># [](start = 3, length = 1)

should we have a section for testing tool guides?

code-reviews/recipes/CSharp.md
```diff
@@ -1,22 +1,20 @@
...
 
-## C# Style Guide
+## Style Guide
```
saroup: Noticed this one doesn't have build validation and dependencies like the bash recipe. Is this intentional?

code-reviews/recipes/Go.md
```diff
@@ -2,47 +2,55 @@
...
-### ```gofmt```
+### go vet
+
+`go vet` is a static analysis tool that checks for common go errors, such as incorrect use of range loop variables or misaligned printf arguments. [CSE](../../CSE.md) Go code should be able to build with no `go vet` errors.
+
+### golint
+
+[golint](https://github.com/golang/lint) can be an effetive tool for finding many issues, but it errors on the side of false positives. It is best used by developers when working on code, not as part of an automated build process.
+
+## Automatic Code Formatting
```
saroup: Is this not the same as linter?

code-reviews/recipes/Go.md
```diff
@@ -2,47 +2,55 @@
...
-[```golint```](https://github.com/golang/lint) can be an effetive tool for finding many issues, but it errors on the side of false positives. It is best used by developers when working on code, not as part of an automated build process.
+`gofmt` should be run as a part of every build to enforce the common standard.
 
+To automate this process in Azure Devops you can add the following snippet to you `azure-pipelines.yaml` file. This will lint any scripts in the `./scripts/` folder.
 
-### Automation
+> TODO: Add example of yaml for gofmt
 
-#### `golangci-lint`
+## Pre-Commit Hooks
```
saroup: This is basically like the previous section . I know you are not modifyin content but I don't think there should be a section for Pre-commits. All the the pre commit work should be part of the other sections. like running linters and tests etc.

code-reviews/recipes/JavaScript.md
```diff
@@ -2,27 +2,136 @@
...
 ## Style Guide
 
-We use [`prettier`](https://prettier.io/) to do code formatting for JavaScript. Using an automated code formatting tool like Prettier enforces a well accepted style guide that was collaboratively built by a wide range of companies including Microsoft, Facebook, and AirBnB. It is highly recommended that you use one of the native editor integrations and enable format on save. In addition, by default, we use the following overrides (shown here in the config format for VSCode) to standardize on single quotes, a four space drop, and to do ESLinting:
+[CSE](../../CSE.md) developers use [prettier](https://prettier.io/) to do code formatting for JavaScript.
 
+Using an automated code formatting tool like Prettier enforces a well accepted style guide that was collaboratively built by a wide range of companies including Microsoft, Facebook, and AirBnB.
+
+For higher level style guidance not covered by prettier, we follow the [AirBnB Style Guide](https://github.com/airbnb/javascript).
+
+## Linter
```
saroup: Maybe this should be called linting instead

code-reviews/recipes/JavaScript.md
```diff
@@ -2,27 +2,136 @@
```
saroup: I am wondering if we should also rename the file to reflect JS & TS

code-reviews/recipes/Bash.md
```diff
@@ -2,15 +2,17 @@
...
 
-[CSE](../../../CSE.md) developers follow [Google's Bash Style Guide](https://google.github.io/styleguide/shell.xml).
+[CSE](../../CSE.md) developers follow [Google's Bash Style Guide](https://google.github.io/styleguide/shell.xml).
 
 ## Linter
 
-[CSE](../../../CSE.md) projects must check Bash code with [shellcheck](https://github.com/koalaman/shellcheck) as part of the [CI process](../../../continuous-integration/readme.md).
+[CSE](../../CSE.md) projects must check Bash code with [shellcheck](https://github.com/koalaman/shellcheck) as part of the [CI process](../../continuous-integration/readme.md).
+
+## Build Validation
```
saroup: When I read build validation, I think that the section will be about how to compile a build for this language. It doesn't necessarily convey that is about linting. or maybe it is not, it just happens that the current content is only linting.
It would be helpful to add a mini description for this section to describe what should go there for anyone writing new recipes?

saroup:
I think these recipes are missing guidelines on testing tools/platforms for each language. When starting out in a new language or in a new organization it is very helpful to know what others are using for testing. what do you think?
I noticed that the recipes don't all have the same sections, Is that intentional? For example, dependencies is not always there. If we think that the recipes should contain relevant info about dependencies it would be nice to have it there and add a TODO to think about any guideline we want to add there.

code-reviews/recipes/CSharp.md
TessFerrandez: If we have that we should have a section for observability and security as well. 
I think the idea with the current play book structure is to have testing covered in the test section.  If we have it in both sections we will have duplicate content and if it is not kept up to date we may end up with conflicting documentation

code-reviews/recipes/CSharp.md
TessFerrandez: I think @debugerr will add that as he updates the C# section.

code-reviews/recipes/Go.md
TessFerrandez: Do you mean golint, or the Automatic Code Formatting?  If you are referring to automatic code formatting, then no.  Code formatting will format the style of the code, maybe adding spaces, removing unused imports etc.  Linting usually looks for if the style is per specification, but often looks for other things as well, like logic errors, missing doc strings etc. like a compiler for interpreted languages.

code-reviews/recipes/Go.md
TessFerrandez: The reason I added a Pre-Commit Hooks section is because this is handled differently in different languages.  

I think the linting section should discuss preferred options for linters, the build validation section discusses how to enforce it as a build validation step to protect the develop branch, and finally the pre-commit hook discusses how to implement the most common optional pre-commit settings.

In python for example, this is most often done with something like the pre-commit package, while in other languages maybe you manually update the git hooks.

I am open for alternatives of course, but it felt pretty natural to leave it as a separate section to avoid cluttering the linting section.

code-reviews/recipes/JavaScript.md
TessFerrandez: Can you elaborate on your reasoning here?

code-reviews/recipes/JavaScript.md
TessFerrandez: Sure, I'll do this when I have gotten responses to all the other comments, and will do all changes at once.

code-reviews/recipes/Bash.md
TessFerrandez: Interesting... 

For AzDO build validation is defined as below.  I thought this was common terminology but perhaps it is not.
https://docs.microsoft.com/en-us/azure/devops/repos/git/branch-policies?view=azure-devops

Where do you suggest this wording would go?  

I am thinking that someone that writes a new recipe would first look at the ones that already have this and follow suit, but perhaps that can't be assumed... I am lost as to where to write it though.

TessFerrandez: Interesting.... I thought this was common terminology, perhaps it is just an AzDO thing
https://docs.microsoft.com/en-us/azure/devops/repos/git/branch-policies?view=azure-devops

I was assuming anyone writing a new recipe would perhaps look at the others and get the gist, but perhaps we should clarify... I am at a bit of a loss though as to where to put it.  Do you have a suggestion? and what it should say?

TessFerrandez:
> I think these recipes are missing guidelines on testing tools/platforms for each language. When starting out in a new language or in a new organization it is very helpful to know what others are using for testing. what do you think?
> I noticed that the recipes don't all have the same sections, Is that intentional? For example, dependencies is not always there. If we think that the recipes should contain relevant info about dependencies it would be nice to have it there and add a TODO to think about any guideline we want to add there.

- See comments about testing above...
- Regarding different sections, some will have language specific sections like dependencies. The common sections that would be relevant to most are listed in the task.  But even those would not be required. As an example, linters will not be applicable to C# or Java

code-reviews/recipes/CSharp.md
saroup: sounds good!

code-reviews/recipes/JavaScript.md
saroup: no specific reason just my personal preference. I was thinking of "Linting Guide"   Instead of Linter or Linters

code-reviews/recipes/Go.md
saroup: I was looking at gofmt that is marked as both linter and automatic code formatting tool. Code formatting as you describe it Is still linting to me. Linters will format the style of the code and fix spaces and remove unused imports (like tslint) . Is the code formatting separate from linters or can it also be part of the linter?

TessFerrandez: ahh, I think in that case tslint is doing both linting and code formatting. most likely golint does too, I am not a go expert so can't say

code-reviews/recipes/Go.md
saroup: again , I think this is also a matter of personal preference (sorry!)  if I go to the linting section and we are suggesting to run gofmt before committing I wouldn't think to come to this section to see how to actually do it I would just search the internet. This is why I was suggesting we add the how in every section respectively. really up to you .

code-reviews/recipes/JavaScript.md
TessFerrandez: Cool, I think linter/s fits better with the theme of the other section titles, but let's see what @debugerr thinks

code-reviews/recipes/Go.md
saroup: It was there before your change and I wasn't sure if you intentionally moved it to code formatting. It makes sense for me and it really depends on the language and tools we are using we can have both in the linter or have them separate. I never really thought about code formatting as separate probably because I never had to use a separate tool for it.

code-reviews/recipes/JavaScript.md
davidkpiano: I like linting because "linting" is something that is done by a "linter" (but can also be done manually), so in terms of categorizing it, "linting" is broader.

code-reviews/recipes/Bash.md
saroup: I think it is fine to have linting as part of the build but I hope we won't just add linting to build validation and rather add how to validate a build by compiling the language specific code (and running tests). The other sections also moslty include validating linting.

saroup: maybe you can add a TODO to have language specific build validation in this section.

code-reviews/recipes/JavaScript.md
TessFerrandez: Fixed

code-reviews/recipes/JavaScript.md
TessFerrandez: Fixed

code-reviews/recipes/CSharp.md
debugerr: Absolutely.

code-reviews/recipes/JavaScript.md
debugerr: Both work for me. I agree linting is a bit broader.

code-reviews/recipes/Bash.md
debugerr: To me build validation is really the process of automatically validating as much as possible of the code that is being PR-ed into the repo. For example, in the project we are running at the moment, we run StyleCop, FxCop, Unit tests and there is even a manual step to trigger an E2E test pipeline. Anything that aids validating the code before we have to manually look at it goes in here in my opinion.

PR #157, opened 2020-02-18
Feature/155 specific git guidance

MS-Santi:
As a new contributor, I found I needed some guidance on how to get started. The modifications in this file are what I found was missing. Definitely not a complete list, but including the basics:

- How to get permissions
- What linter to use
- What is this repo's branch and merge strategy.


CONTRIBUTING.md
```diff
@@ -2,7 +2,7 @@
 
 This project welcomes contributions and suggestions.  Most contributions require you to agree to a
 Contributor License Agreement (CLA) declaring that you have the right to, and actually do, grant us
-the rights to use your contribution. For details, visit https://cla.microsoft.com.
+the rights to use your contribution. For details, visit <https://cla.microsoft.com>.
```
davidkpiano: Why the brackets?

CONTRIBUTING.md
```diff
@@ -12,9 +12,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+## Permissions & Contributions
 
-##Structure/Pattern
+There are two ways in which you can help update the content:
+
+* **One-off:** \
+If you are not a regular contributor to the project, but you would like to contribute some changes, the best way to do it is:
+
+  1. Fork the repo
+  1. Create a PR
```
TessFerrandez: Nit: should we use 
1. ...
2. ...

CONTRIBUTING.md
```diff
@@ -12,9 +12,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+* **Periodic and regular contributions:** \
+If you plan to update the content semi-regularly or regularly, you can be added to the project's Contributors group. Please contact one of the project's administrators to be added to the group.
+
+## Git guidance
+
+Consistent with the practices suggested in this playbook, please follow the specifics regarding git as described in this section.
+
+### Branch naming convention
+
+In this repo, we will the following branch naming conventions:
```
TessFerrandez: we will -> we use

CONTRIBUTING.md
```diff
@@ -12,9 +12,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+| Personal | users/\<alias>/\<short description> | johndo/rewrite-grooming-section
+
+> **Note:**
+>
+> * Use of personal branches should be minimized since most work should refer back to a feature/bug fix in the backlog.
+> * Mind the capitalization of the branch prefix (feature, fix, users). Tools that diplay branches as a hierarchy are typically case sensitive, and will display different hierarchies for the same words with different capitalization.
+
+### Linting
+
+Please install the [markdownlint extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) and ensure that all rules are followed. This will help ensure consistency in the look and feel of the documentation in this repo.
```
TessFerrandez: should probably mention that this is for VS Code

CONTRIBUTING.md
```diff
@@ -12,9 +12,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+
+### Branch naming convention
+
+In this repo, we will the following branch naming conventions:
+
+| Branch Type       | Pattern     | Example     |
+| ------------- | ---------- | ----------- |
+|  Feature | feature/\<issue#>-\<short description> | feature/498-reorganize-scm-section |
+| Bug Fix | fix/\<bug#>-\<short description> | bug/978-correct-grammar-myfile.md |
+| Personal | users/\<alias>/\<short description> | johndo/rewrite-grooming-section
```
TessFerrandez: To make it simple, we could just skip the personal branches and the recommendation to not use personal branches

CONTRIBUTING.md
```diff
@@ -12,9 +12,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+> * Mind the capitalization of the branch prefix (feature, fix, users). Tools that diplay branches as a hierarchy are typically case sensitive, and will display different hierarchies for the same words with different capitalization.
+
+### Linting
+
+Please install the [markdownlint extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) and ensure that all rules are followed. This will help ensure consistency in the look and feel of the documentation in this repo.
+
+### Merging strategy
+
+A merging strategy has not been formally adopted.
+For the time being, and until one is adopted, familiarize yourself with the [git guidance in this repo](https://github.com/microsoft/code-with-engineering-playbook/blob/master/source-control/git.md), decide whether you'll favor liner or non-linear commit history and follow that procedure.
```
TessFerrandez: Nit: whether you'll favor liner  [spelling on you'll and liner] -- whether you favor linear

CONTRIBUTING.md
```diff
@@ -12,9 +12,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+> **Note:**
+>
+> * Use of personal branches should be minimized since most work should refer back to a feature/bug fix in the backlog.
+> * Mind the capitalization of the branch prefix (feature, fix, users). Tools that diplay branches as a hierarchy are typically case sensitive, and will display different hierarchies for the same words with different capitalization.
+
+### Linting
+
+Please install the [markdownlint extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) and ensure that all rules are followed. This will help ensure consistency in the look and feel of the documentation in this repo.
+
+### Merging strategy
```
TessFerrandez: Since we haven't decided on a merging strategy, either decide on one, (I prefer non-linear), or remove the section, or add a TODO to decide. As is, this section doesn't add anything to the document

CONTRIBUTING.md
```diff
@@ -12,9 +12,54 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+
+A merging strategy has not been formally adopted.
+For the time being, and until one is adopted, familiarize yourself with the [git guidance in this repo](https://github.com/microsoft/code-with-engineering-playbook/blob/master/source-control/git.md), decide whether you'll favor liner or non-linear commit history and follow that procedure.
+
+### Releases
```
TessFerrandez: Same here, this section does not add to the document, my suggestion is to remove it.
There is no release, so no need for a release strategy

CONTRIBUTING.md
```diff
@@ -66,9 +111,9 @@ see the [LICENSE](LICENSE) file, and grant you a license to any code in the repo
 Microsoft, Windows, Microsoft Azure and/or other Microsoft products and services referenced in the documentation
 may be either trademarks or registered trademarks of Microsoft in the United States and/or other countries.
 The licenses for this project do not grant you rights to use any Microsoft names, logos, or trademarks.
-Microsoft's general trademark guidelines can be found at http://go.microsoft.com/fwlink/?LinkID=254653.
+Microsoft's general trademark guidelines can be found at <http://go.microsoft.com/fwlink/?LinkID=254653>.
```
TessFerrandez: Same question as above regarding the brackets

CONTRIBUTING.md
```diff
@@ -66,9 +111,9 @@ see the [LICENSE](LICENSE) file, and grant you a license to any code in the repo
...
 
-Privacy information can be found at https://privacy.microsoft.com/en-us/
+Privacy information can be found at <https://privacy.microsoft.com/en-us/>
```
TessFerrandez: and again...

CONTRIBUTING.md
MS-Santi: Part of the markdown rules are not to have "bare" URLs. Thus, you indicate that the URL is explicitly typed and it should not be text or have an alternative text.

CONTRIBUTING.md
MS-Santi: See above - honors MD034/no-bare-urls: Bare URL

CONTRIBUTING.md
MS-Santi: See above - honors MD034/no-bare-urls: Bare URL

CONTRIBUTING.md
MS-Santi: I like it - I just added it because I saw a couple of them, and they are mentioned in the guidance. I removed and modified the note

CONTRIBUTING.md
TessFerrandez: Correction: Looks like the default is squash and commit, and that most people so far has done this so we should probably change this to linear

MS-Santi: I have added the task in the backlog - I'll change appropriately.

CONTRIBUTING.md
```diff
@@ -12,9 +12,49 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
...
+* **Periodic and regular contributions:** \
+If you plan to update the content semi-regularly or regularly, you can be added to the project's Contributors group. Please contact one of the project's administrators to be added to the group.
+
+## Git guidance
+
+Consistent with the practices suggested in this playbook, please follow the specifics regarding git as described in this section.
+
+### Branch naming convention
+
+In this repo, we use the following branch naming conventions:
```
rozele: I think we should disallow branches other than master in this repo, and request that people use their own forks to make pull requests.

rozele: Thoughts @TessFerrandez?

TessFerrandez: I think that sounds good, especially as it is an OSS repo, but then I think it needs to be enforced as @rozele suggests to avoid the current branch situation

MS-Santi: I'll change the text to describe this approach. @TessFerrandez, @rozele - who has permissions to configure the repo accordingly?

fnocera: @MS-Santi @rozele @TessFerrandez I can change the policies on the repo if this is what we have decided. However, I am hesitating a little bit as although branches can clutter the repo (we can fix this by having an automated delete once a branch is merged) I think forks create much less visibility for what work is in progress as well as higher friction for contributions. Thoughts?

TessFerrandez: @fnocera I was thinking the same today actually, as I was looking at some PRs. If you want to clone it locally to take it through the linter for example you have to create a separate folder, it was much more cumbersome than switching branches.

I would like to hear @rozele’s thoughts though. I was agreeing with him as it seemed like a good way to avoid all the leftover branches, but if we can add an automatic delete, that seems to solve that problem.  

If we keep the branches, I think it would be a good idea to have a common naming policy to understand what the branch is about.

TessFerrandez: TBH I don't have a strong opinion on this, I think your comment regarding friction for contributors is very valid and perhaps that is the most important thing at this point

fnocera: @TessFerrandez @rozele @MS-Santi 
Eric, Tess and I had a discussion today and for the near term (next few months) we will not disable push permissions/force people to use forks. Once the number of issues/contributions to the playbook have leveled out we can move to a fork only model.  Therefore we do need branch naming conventions!

MS-Santi: Sounds good. I'll go back to the version with the naming conventions

MS-Santi:
@TessFerrandez 

> No need to change anything if you don't agree, but just a quick thought... do linters even make sense with a compiled language? or should we remove this wording, alternatively change the title to Code Analyzers or similar?

I think that we are using a term that is well known -linter-. In addition, a compiler doesn't really care if you put the opening brace on the same line or the next one. I believe that for compiled languages, the linter will mostly help with the stylistic checks, whereas for interpreted languages, the linter will go deeper to find more profound problems.

TessFerrandez:
> See comment, I actually think we should disable all push permissions on this repository and force people to use forks. At that point, I think we don't really need to worry about how people name branches on their forks.

@rozele can you resolve this comment, per our earlier conversation

PR #181, opened 2020-03-03
Adding markdownlint as a github action for our repo

### LINKED ISSUE #180: Repo should impose markdown linting
Create a github workflow/action that runs markdown linting on Pull Requests. 
eg: Something like https://github.com/dotnet/docs/blob/821a0e0b7e688fcdaf70276450f9858359919888/.github/workflows/markdownlint.yml

Disable: 
- Line length limit

fnocera:
Resolves #180 by adding a github workflow/action to run linting on PRs to mater.


.markdownlint.json
```diff
@@ -0,0 +1,12 @@
+{
```
rozele: Why use `.markdownlint.json` instead of `.markdownlintrc`? Just curious.

.markdownlint.json
```diff
@@ -0,0 +1,12 @@
...
+  "default": true,
+  "MD004": false,
```
rozele: Why not just enable this and change all the `*` to `-`, it's a pretty trivial fix. Would drive consistency for unordered lists.

.markdownlint.json
```diff
@@ -0,0 +1,12 @@
...
+  "MD013": false,
```
rozele: nit / opinion - the descriptive names (e.g., "line-length" in this case) are easier to read / understand. Using the rule IDs requires opening the map.

.markdownlint.json
```diff
@@ -0,0 +1,12 @@
...
+  "MD025": false,
```
rozele: This one could be enabled. Only `CONTRIBUTING.md` uses this, and we can easily nest the inner headings.

.markdownlint.json
```diff
@@ -0,0 +1,12 @@
...
+  "MD026": false,
+  "MD029": false,
```
rozele: This is another good rule to keep. There's only one place where this was violated, and it was due to a missing newline between the list sections, not an actual violation.

.markdownlint.json
```diff
@@ -0,0 +1,12 @@
...
+  "MD036": false,
```
rozele: Another decent idea. Headings at least give link anchors. The only place where this was used was `CICD.md` and the emphasized list items can easily be converted to headings.

Issue #88, opened 2020-01-21
Tooling: Add Visual Studio & VS Code guidance


debugerr:
Add guidance on how to optimize the CR experience using Visual Studio and VS Code to support an easier review process vs. using the web UI of git.com

1. supported plugins
2. step by step setup guide
3. tips and tricks

PR #190, opened 2020-03-04
Adding tools section for reviewing PRs

debugerr:
Adding a section in the readme.md to link to a tools section. Tools can be anything that support the code review process.

I've added tools for VS Code and VS. All of them support the review process out of the VS family and allow to checkout and navigate the code easier than when using the online tools provided by ADO or GitHub.

This is closing #88.

### DIFF
--- a/code-reviews/tools.md
+++ b/code-reviews/tools.md
@@ -1,0 +1,34 @@
+# Code review tools
+
+## Visual Studio Code
+
+### GitHub: [GitHub Pull Requests](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github)
+
+Supports GitHub pull request processing inside of VS Code.
+
+1. Open the plugin from the "Activity Bar"
+1. Select "Assigned To Me"
+1. Select a PR
+1. Under "Description" you can chose to "Check Out" the branch and get into "Review Mode" and get a more integrated experience
+
+## Visual Studio
+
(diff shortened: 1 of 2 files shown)


code-reviews/tools.md
```diff
@@ -0,0 +1,34 @@
+# Code review tools
+
+## Visual Studio Code
+
+### GitHub: [GitHub Pull Requests](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github)
+
+Supports GitHub pull request processing inside of VS Code.
```
tompaana: How about "Supports processing GitHub pull requests inside VS Code."

code-reviews/tools.md
```diff
@@ -0,0 +1,34 @@
...
+
+1. Open the plugin from the "Activity Bar"
```
tompaana: Matter of taste and just a thought: When referring to items in menus etc. rather than using quotation marks, one could use different font style (italic/bold)

code-reviews/tools.md
debugerr: fixed

code-reviews/tools.md
debugerr: I don't really mind. We should try to be consistent. Let me try to find some other occurences.

debugerr: Azure docs is using bold as an indication for Menus/Links. I think I'll go for that. Thanks.

PR #195, opened 2020-03-04
Obs: Adding section on Pitfalls/what to avoid

NarmathaBala:


observability/pitfalls.md
```diff
@@ -0,0 +1,20 @@
+# What to watch for when building Observable Systems
+
+## Observability as an after thought
+
+One of the design goals when building a system must be to have a monitorable system. This helps planning and thinking application availability, logging and metrics at the time of design and development. Observability also acts as a great debugging tool providing developers a bird's eye view of the system. By leaving instrumentation and logging of metrics towards the end, the development teams lose valuable insights during development.
+
+## Metric Fatigue
+
+It is recommended to collect and measure *what you want* **and** *not what you can*. Don't attempt to monitor everything. If the data is not actionable, it is useless and becomes noise. On the contrary, it is sometimes very difficult to forecast every possible scenario that could go wrong. There is a balance between collecting what is needed and logging every single activity in the system. A general rule of thumb is to follow these principles
```
TessFerrandez: Suggestion: consider changing to a list to make the text more readable as there is a lot of good and important information here

khaines: I would also suggest changing `what you want` to `what you need`, because what is needed for measuring may not be what the operator thinks is the most interesting to measure.

README.md
```diff
@@ -45,7 +45,7 @@ A [breakdown of sections](SPRINT-STRUCTURE.md) according to the structure of an
 * [CI](continuous-integration/readme.md)
 * [CD](continuous-deployment/readme.md)
 * [Code Reviews](code-reviews/readme.md)
-* [Observability](observability/readme.md)
+* [Observability](observability/README.md)
```
fnocera: We have all readme's lower case so I would keep it consistent if possible - ie. Rename observability/README.md

PR #200, opened 2020-03-05
Add git LFS guidance

TessFerrandez:
- adds guidance on how to deal with large binary files
- pros and cons of Git LFS
- common commands for Git LFS
- references to more Git LFS info

### DIFF
--- a/source-control/git.md
+++ b/source-control/git.md
@@ -279,3 +279,9 @@ Download files to your working directory
 git lfs pull
 git lfs pull --include="path/to/file"
 ```
+
+#### References
+
+* [Git LFS getting started](https://git-lfs.github.com/)
+* [Git LFS manual](https://github.com/git-lfs/git-lfs/tree/master/docs)
+* [Git LFS on Azure Repos](https://docs.microsoft.com/en-us/azure/devops/repos/git/manage-large-files?view=azure-devops)
--- a/source-control/git.md
+++ b/source-control/git.md
@@ -236,7 +236,7 @@ One solution to this problem is `Git LFS (Git Large File Storage)` - an open sou
   * Binary files committed through Git LFS are not visible as Git will only download the data describing the large file
   * Committing large binaries will push the full binary to the repository
 * Git cannot merge the changes from two different versions of a binary file; file locking mitigates this
-* Azure Repos do not support using SSH for repositories with Git LFS tracked files
+* Azure Repos do not support using SSH for repositories with Git LFS tracked files - for more information see the Git LFS [authentication documentation](https://github.com/git-lfs/git-lfs/blob/master/docs/api/authentication.md)
 
 #### Common commands
 
@@ -247,13 +247,15 @@ git lfs install       # windows
 sudo apt-get git-lfs  # linux
 ```
 
+See the [Git LFS installation instructions](https://github.com/git-lfs/git-lfs/wiki/Installation) for installation on other systems
+
 Track .mp4 files with Git LFS
 
 ```bash
 git lfs track '*.mp4'
 ```
 
-Add a `.gitattributes` file listing the files and patterns to track
+Update the `.gitattributes` file listing the files and patterns to track
 
 ```bash
 *.mp4 filter=lfs diff=lfs merge=lfs -text
--- a/source-control/git.md
+++ b/source-control/git.md
@@ -184,7 +184,7 @@ git push origin HEAD --force
 
 ### Multi service development and deployments
 
-Submodules can be very useful in more complex deployment and/or development scenarios
+Submodules can be useful in more complex deployment and/or development scenarios
 
 Adding a submodule to your repo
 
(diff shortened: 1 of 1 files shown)


source-control/git.md
```diff
@@ -211,8 +211,77 @@ git stash
...
-Or you can simply move the current state into a new branch:
+Or you can move the current state into a new branch:
 
 ```bash
 git stash branch <new_branch_to_save_changes>
 ```
+
+### Working with images, video and other binary content
+
+Avoid committing frequently changed binary files, such as large images, video or compiled code to your git repository. Binary content is not diffed like text content, so the repository will take longer and longer to clone.
```
rozele: Great job removing superlatives above. Can simplify `longer and longer` to just `longer`, or make this more concrete by saying "Binary content is not diffed like text content, so cloning or pulling from the repository may pull each revision of the binary file".

source-control/git.md
```diff
@@ -211,8 +211,77 @@ git stash
...
+
+One solution to this problem is `Git LFS (Git Large File Storage)` - an open source Git extension for versioning large files. `Git LFS` replaces the binary files with text pointers inside Git, while storing the file contents on a remote server like GitHub.com or Azure DevOps.
+
+#### Benefits of Git LFS
+
+* Uses the end to end Git workflow for all files
+* Binary files tracked by Git LFS can be as big as you need them to be
```
rozele: What does this statement mean? There is no upper limit on the size of files tracked by Git LFS? Not sure that is true: https://github.com/git-lfs/git-lfs/wiki/Limitations

source-control/git.md
```diff
@@ -211,8 +211,77 @@ git stash
...
+* Git LFS supports file locking to help the team collaborate on large, undiffable assets
```
rozele: As a reader (and someone unfamiliar with git LFS), curious how file locking supports collaboration.

source-control/git.md
```diff
@@ -211,8 +211,77 @@ git stash
...
+* Git LFS is fully supported in Azure DevOps Services
+
+#### Limitations of Git LFS
+
+* Everyone contributing to the repository needs to install Git LFS
+* If not set up properly:
+  * Binary files committed through Git LFS are not visible as Git will only download the data describing the large file
+  * Committing large binaries will push the full binary to the repository
+* Git cannot merge the changes from two different versions of a binary file. File locking mitiages this
```
rozele: `mitiages` -> `mitigates`

rozele: (nitpick) Odd to have mix of complete sentences and not in a list like this. Maybe consider - `... versions of a binary file; file locking mitigates this`

source-control/git.md
TessFerrandez: I'll remove that statement, I believe you're right

source-control/git.md
TessFerrandez: I'll think about how to rephrase that...

source-control/git.md
```diff
@@ -211,8 +211,76 @@ git stash
...
 
 ```bash
 git stash branch <new_branch_to_save_changes>
 ```
+
+### Working with images, video and other binary content
+
+Avoid committing frequently changed binary files, such as large images, video or compiled code to your git repository. Binary content is not diffed like text content, so cloning or pulling from the repository may pull each revision of the binary file.
+
+One solution to this problem is `Git LFS (Git Large File Storage)` - an open source Git extension for versioning large files. `Git LFS` replaces the binary files with text pointers inside Git, while storing the file contents on a remote server like GitHub.com or Azure DevOps.
```
anthturner: Nit: One thing we may want to include here is a note about how Git LFS influences security boundaries of code/data in GitOps. Since files are not stored in the repository, it's worth clarifying how that might affect customers who are particularly sensitive to issues like source code leak and developer asset theft.

Linking to something like this would be helpful for people dealing with those more data-boundary-aware customers: https://github.com/git-lfs/git-lfs/blob/master/docs/api/authentication.md

(This is especially important given that you note a couple sections down that Azure Repos don't support SSH connections for Git LFS)

source-control/git.md
```diff
@@ -211,8 +211,76 @@ git stash
...
+* Git cannot merge the changes from two different versions of a binary file; file locking mitigates this
+* Azure Repos do not support using SSH for repositories with Git LFS tracked files
+
+#### Common commands
+
+Install git lfs
+
+```bash
+git lfs install       # windows
+sudo apt-get git-lfs  # linux
```
anthturner: Nit: May want to quickly look up what the equivalent packages are for `yum` (RedHat/Fedora/CentOS and variants) and `pacman` (ArchLinux and variants) because not everyone will be using Debian/Ubuntu. I think those 3 cover most of the package management for any Linux server distros I'd expect to see currently in the wild.

source-control/git.md
```diff
@@ -211,8 +211,76 @@ git stash
...
+```
+
+Track .mp4 files with Git LFS
+
+```bash
+git lfs track '*.mp4'
+```
+
+Add a `.gitattributes` file listing the files and patterns to track
```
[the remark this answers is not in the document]
TessFerrandez: I will change the wording

source-control/git.md
TessFerrandez: Fixed

source-control/git.md
TessFerrandez: Fixed

source-control/git.md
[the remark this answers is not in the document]
TessFerrandez: Fixed

Issue #131, opened 2020-02-11
Develop design review template for features/stories


wbreza:

manojvazirani:
Closing this as done as merged into master with PR #280

PR #239, opened 2020-03-17
Add task level design review template to design reviews templates

manojvazirani:
Added task level design review template based on @wbreza prior work and incorporating some other ideas based on prior experience.

#132

### DIFF
--- a/design-reviews/recipes/task-design-review-template.md
+++ b/design-reviews/recipes/task-design-review-template.md
@@ -13,38 +13,38 @@
 
 ## Overview / Problem Statement
 
-- Describe the task with a high-level summary. 
-- Consider additional background and justification, for posterity and historical context.
+* Describe the task with a high-level summary.
+* Consider additional background and justification, for posterity and historical context.
 
 ## Goals / In-Scope
 
-- List the specific goals for the task.
-- List a few bullet points of what this task will achieve and that are most relevant for the design review discussion.
-- This should include acceptance criteria required to meet the [definition of done](../../team-agreements/definition-of-done/).
+* List the specific goals for the task.
+* List a few bullet points of what this task will achieve and that are most relevant for the design review discussion.
+* This should include acceptance criteria required to meet the [definition of done](../../team-agreements/definition-of-done/).
 
 ## Non-goals / Out-of-Scope
 
-- List the non-goals for the task.  
-- List a few bullet points of non-goals to clarify the work that is beyond the scope of the design review for this task.
+* List the non-goals for the task.  
+* List a few bullet points of non-goals to clarify the work that is beyond the scope of the design review for this task.
 
 ## Proposed Design / Suggested Approach
 
-- Describe the detailed design to accomplish the proposed task. 
-- What patterns & practices will be used and why were they chosen.  
-- Were any alternate proposals considered?  
-- What new components are required to be developed? 
-- Are there any existing components that require updates?
-- Relevant diagrams (e.g. sequence, component, context, deployment) should be included here.
+* Describe the detailed design to accomplish the proposed task.
+* What patterns & practices will be used and why were they chosen.  
+* Were any alternate proposals considered?  
+* What new components are required to be developed?
+* Are there any existing components that require updates?
+* Relevant diagrams (e.g. sequence, component, context, deployment) should be included here.
 
-## Technology Choices 
+## Technology Choices
 
-- Describe any libraries and OSS components that will be used to complete the task. 
-- Briefly list the languages(s) and platform(s) that comprise the stack.
+* Describe any libraries and OSS components that will be used to complete the task.
+* Briefly list the languages(s) and platform(s) that comprise the stack.
 
-## Open Questions 
+## Open Questions
 
-> List any open questions/concerns here. 
+> List any open questions/concerns here.
 
-## Additional References 
+## Additional References
 
 > List any additional references here including links to backlog items, work items or other documents.
--- a/design-reviews/recipes/task-design-review-template.md
+++ b/design-reviews/recipes/task-design-review-template.md
@@ -1,0 +1,50 @@
+# Your Task Design Title Here (prefix with DRAFT/WIP to indicate level of completeness)
+
+> When developing a design document for a new task it should contain a detailed design proposal demonstrating how it will solve the goals outlined below.  
+> Not all tasks require a design review but when they do it is likely that there many unknowns or the solution may be more complex.
+> The design should include diagrams, pseudo code, interface contracts as needed to provide a detailed understanding of the proposal.
+
+* [Task Name](http://link-to-task-work-item)
+* [Story Name](http://link-to-story-work-item)
+* Organization: [Organization]
+* Engagement: [Engagement]
+* Customer: [Customer]
+* Authors: [Author1, Author2, etc.]
+
+## Overview / Problem Statement
+
+- Describe the task with a high-level summary. 
+- Consider additional background and justification, for posterity and historical context.
+
+## Goals / In-Scope
+
+- List the specific goals for the task.
+- List a few bullet points of what this task will achieve and that are most relevant for the design review discussion.
+- This should include acceptance criteria required to meet the [definition of done](../../team-agreements/definition-of-done/).
+
+## Non-goals / Out-of-Scope
+
+- List the non-goals for the task.  
+- List a few bullet points of non-goals to clarify the work that is beyond the scope of the design review for this task.
+
+## Proposed Design / Suggested Approach
+
+- Describe the detailed design to accomplish the proposed task. 
+- What patterns & practices will be used and why were they chosen.  
+- Were any alternate proposals considered?  
+- What new components are required to be developed? 
+- Are there any existing components that require updates?
+- Relevant diagrams (e.g. sequence, component, context, deployment) should be included here.
+
+## Technology Choices 
+
+- Describe any libraries and OSS components that will be used to complete the task. 
+- Briefly list the languages(s) and platform(s) that comprise the stack.
+
+## Open Questions 
+
+> List any open questions/concerns here. 
+
+## Additional References 
+
+> List any additional references here including links to backlog items, work items or other documents.
--- a/design-reviews/recipes/task-design-review-template.md
+++ b/design-reviews/recipes/task-design-review-template.md
@@ -1,7 +1,7 @@
 # Your Task Design Title Here (prefix with DRAFT/WIP to indicate level of completeness)
 
-> When developing a design document for a new task it should contain a detailed design proposal demonstrating how it will solve the goals outlined below.  
-> Not all tasks require a design review but when they do it is likely that there many unknowns or the solution may be more complex.
+> When developing a design document for a new task, it should contain a detailed design proposal demonstrating how it will solve the goals outlined below.  
+> Not all tasks require a design review, but when they do it is likely that there many unknowns or the solution may be more complex.
 > The design should include diagrams, pseudo code, interface contracts as needed to provide a detailed understanding of the proposal.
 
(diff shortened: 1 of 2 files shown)


design-reviews/recipes/task-design-review-template.md
```diff
@@ -0,0 +1,50 @@
...
+* What patterns & practices will be used and why were they chosen.  
+* Were any alternate proposals considered?  
+* What new components are required to be developed?
+* Are there any existing components that require updates?
+* Relevant diagrams (e.g. sequence, component, context, deployment) should be included here.
+
+## Technology Choices
+
+* Describe any libraries and OSS components that will be used to complete the task.
+* Briefly list the languages(s) and platform(s) that comprise the stack.
```
fnocera: How about including Alternatives Considered also as a separate section? (just after Proposed Design) Breaks the template down a little more.

manojvazirani: Thats a great point @fnocera. They are usually covered under the proposed design section:
`Were any alternate proposals considered?`

design-reviews/recipes/task-design-review-template.md
```diff
@@ -0,0 +1,50 @@
+# Your Task Design Title Here (prefix with DRAFT/WIP to indicate level of completeness)
+
+> When developing a design document for a new task it should contain a detailed design proposal demonstrating how it will solve the goals outlined below.
```
arsenvlad: new task it -> new task, it

design-reviews/recipes/task-design-review-template.md
```diff
@@ -0,0 +1,50 @@
+# Your Task Design Title Here (prefix with DRAFT/WIP to indicate level of completeness)
+
+> When developing a design document for a new task it should contain a detailed design proposal demonstrating how it will solve the goals outlined below.  
+> Not all tasks require a design review but when they do it is likely that there many unknowns or the solution may be more complex.
```
arsenvlad: a design review but when -> a design review, but when

design-reviews/recipes/task-design-review-template.md
```diff
@@ -0,0 +1,50 @@
...
+> The design should include diagrams, pseudo code, interface contracts as needed to provide a detailed understanding of the proposal.
+
+* [Task Name](http://link-to-task-work-item)
+* [Story Name](http://link-to-story-work-item)
```
arsenvlad: Since task is usually nested under a story, do we need both links or just the task is sufficient?

design-reviews/recipes/task-design-review-template.md
```diff
@@ -0,0 +1,50 @@
...
+* Organization: [Organization]
+* Engagement: [Engagement]
+* Customer: [Customer]
```
arsenvlad: Is Customer different from Organization? Maybe just keeping one link for Project / Engagement: ?

design-reviews/recipes/task-design-review-template.md
```diff
@@ -0,0 +1,50 @@
...
+* Authors: [Author1, Author2, etc.]
+
+## Overview / Problem Statement
+
+* Describe the task with a high-level summary.
+* Consider additional background and justification, for posterity and historical context.
+
+## Goals / In-Scope
+
+* List the specific goals for the task.
```
arsenvlad: Since the actual task goals are likely in the work item tracking system, maybe OK to skip this first bullet point?

design-reviews/recipes/task-design-review-template.md
```diff
@@ -0,0 +1,50 @@
...
+* List a few bullet points of what this task will achieve and that are most relevant for the design review discussion.
+* This should include acceptance criteria required to meet the [definition of done](../../team-agreements/definition-of-done/).
+
+## Non-goals / Out-of-Scope
+
+* List the non-goals for the task.
```
arsenvlad: Would 2nd bullet point be sufficient?

design-reviews/recipes/task-design-review-template.md
manojvazirani: It makes sense in ADO but I have seen instances in github where tasks don't have a story mentioned as parent. This will help keep track of the same.

design-reviews/recipes/task-design-review-template.md
manojvazirani: Yeah after reading it couple of times, it seemed to be repetitive.

design-reviews/recipes/task-design-review-template.md
```diff
@@ -0,0 +1,47 @@
...
+## Goals / In-Scope
+
+* List a few bullet points of what this task will achieve and that are most relevant for the design review discussion.
+* This should include acceptance criteria required to meet the [definition of done](../../team-agreements/definition-of-done/).
+
+## Non-goals / Out-of-Scope
+
+* List a few bullet points of non-goals to clarify the work that is beyond the scope of the design review for this task.
+
+## Proposed Design / Suggested Approach
```
charleszipp: I would change the headings to guide the author to fill out a very short description of Options. The reality is if we are doing a design at a task level, we just need to make a decision (not complete a spec and fill out every angle). Since we would typically aim to finish a story in a sprint (1 week), the task needs to be finished in a couple of days. Therefore, I would suggest we focus on people making a decision using summarized information. I would want this to be something the author can finish in a few hours (definitely less than a day). Thoughts @fnocera @arsenvlad ? 

So for example...

## Proposed Options

### Option 1: [`Insert Option Title Here`]
[`TODO: Provide short (2-3 sentances) description how the proposal achieves the goals`]
[`TODO: Provide illustration of the solution (i.e. architecture diagram, flowchart, or sequence). Use minimum needed to convey the suggested change. This should indicate the any changes in the tech stack`]

#### Advantages
[`TODO: Quick bullet list (3 max) what advantages this has over other options`]
#### Risks
[`TODO: Quick bullet list (3 max) what risks if any this has when compared with other options`]

### Option 2: [`Insert Options Title Here`]
[`TODO: complete using same template outlined in option 1`]

design-reviews/recipes/task-design-review-template.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Overview / Problem Statement
+
+* Describe the task with a high-level summary.
+* Consider additional background and justification, for posterity and historical context.
+
+## Goals / In-Scope
+
+* List a few bullet points of what this task will achieve and that are most relevant for the design review discussion.
+* This should include acceptance criteria required to meet the [definition of done](../../team-agreements/definition-of-done/).
```
charleszipp: I disagree, definition of done should be documented as part of the working agreement and apply to all work. I would not duplicate the documentation of that in other areas. I would remove this bullet

design-reviews/recipes/task-design-review-template.md
```diff
@@ -0,0 +1,47 @@
...
+* [Task Name](http://link-to-task-work-item)
+* [Story Name](http://link-to-story-work-item)
+* Engagement: [Engagement]
+* Customer: [Customer]
+* Authors: [Author1, Author2, etc.]
+
+## Overview / Problem Statement
+
+* Describe the task with a high-level summary.
+* Consider additional background and justification, for posterity and historical context.
```
charleszipp: Could this be replaced with a link back to the user story? The user story should capture the what is needed and the justification. I would not want to have two sources of truth for that. So maybe remove this bullet and add to the placeholder's one for the user story?

design-reviews/recipes/task-design-review-template.md
```diff
@@ -0,0 +1,47 @@
...
+> Not all tasks require a design review, but when they do it is likely that there many unknowns or the solution may be more complex.
+> The design should include diagrams, pseudo code, interface contracts as needed to provide a detailed understanding of the proposal.
+
+* [Task Name](http://link-to-task-work-item)
+* [Story Name](http://link-to-story-work-item)
+* Engagement: [Engagement]
+* Customer: [Customer]
+* Authors: [Author1, Author2, etc.]
+
+## Overview / Problem Statement
```
charleszipp: Could this be replaced with a link back to the task? The task should have a title and description and be the source of truth for describing the work.

design-reviews/recipes/task-design-review-template.md
```diff
@@ -0,0 +1,47 @@
...
+
+* Describe the task with a high-level summary.
+* Consider additional background and justification, for posterity and historical context.
+
+## Goals / In-Scope
```
charleszipp: I would repurpose this section to be put more terms of a directional decision that needs to be made. For example... 

## Decision
In order to complete this task, we need to make a decision on the api contract for the new POST endpoint being added with this task. The contract must provide to the consuming service

### Scope
* An identifier that can be used to look up the created resource
* Status indicator if the downstream processing completed

I might templatize that as...

## Decision
[`TODO: Summarize the decision that needs to be made before we can complete this task`]

### Scope
[`TODO: Add bullet list of what this task must fulfill to make the story as a whole pass acceptance`]

Thoughts @fnocera @arsenvlad

design-reviews/recipes/task-design-review-template.md
fnocera: We don't have a template for user stories, am I correct? I would actually rename this whole contribution to be a template for user stories. In my experience we don't do design for tasks and if we do it is an incremental addition to the design document for the user story or would agree with Charles' suggestion above. 
Thoughts @charleszipp @arsenvlad @manojvazirani

design-reviews/recipes/task-design-review-template.md
wbreza: Agreed - we should include a link to the source task in whichever repo/tool they are using.

design-reviews/recipes/task-design-review-template.md
wbreza: I agree, but at the same time i feel it is important for historical purposes.  Think of the scenario where this design review could be leveraged by a future engagement and we no longer have access to the user story in the customers account.

design-reviews/recipes/task-design-review-template.md
wbreza: @fnocera FYI - The template for using stories is in our backlog and coming soon.

design-reviews/recipes/task-design-review-template.md
wbreza: Similar to my note above.  I don't think we should be duplicating data available from othe sources but also feel the design review should have enough information to review without having to reference other material so that it can easily be referenced at a later point in time after a project goes out of scope.

design-reviews/recipes/task-design-review-template.md
manojvazirani: @fnocera : #131 is the task for creating template for user stories.

I have updated the template to include proposed options which will be helpful to summarize task level decisions.

design-reviews/recipes/task-design-review-template.md
manojvazirani: I have added the flexible option to link to the task so we can have the details even if the link is unavailable.

design-reviews/recipes/task-design-review-template.md
manojvazirani: I have added the flexible option to link to the task so we can have the details even if the link is unavailable.

design-reviews/recipes/task-design-review-template.md
manojvazirani: I agree with the suggestion to add 'decision', but it seemed repetitive to the proposed options and scope would clearly define the goal, so had to lean towards the previous template over this approach.

design-reviews/recipes/task-design-review-template.md
manojvazirani: I agree with @wbreza on the fact that there should be enough information on the document so it can be reviewed as an independent document without having to reference other material.

PR #244, opened 2020-03-18
Restructuring obs/readme.md

### LINKED ISSUE #240: Enhancement: Reorg observability readme for readability
1. Adding table of contents to readme.md
2. Organizing content for readability

NarmathaBala:
1. Added TOC for better readability
2. Removed 'What to collect" section as it is now redundant

closes #240

### DIFF
--- a/observability/readme.md
+++ b/observability/readme.md
@@ -2,6 +2,22 @@
 
 Building observable systems enables development teams at CSE to measure how well the application is behaving.
 
+## Table of Contents
+
+- [Observability](#observability)
+  - [Table of Contents](#table-of-contents)
+  - [Goals](#goals)
+  - [Pillars of Observability](#pillars-of-observability)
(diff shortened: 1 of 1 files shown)


observability/readme.md
```diff
@@ -2,6 +2,22 @@
 
 Building observable systems enables development teams at CSE to measure how well the application is behaving.
 
+## Table of Contents
```
TessFerrandez: Should we start adding these TOCs more consistently?
In source code and code reviews we follow this pattern https://github.com/microsoft/code-with-engineering-playbook/tree/master/code-reviews - i.e. **Table of Contents** is bolded and not a title, and only one level is visible in the TOC - level 2 titles.

If you prefer this way, we should maybe start a conversation in the teams channel so that we can make all teams conform to whatever decision is made

PR #245, opened 2020-03-18
Adding Java Code Review recipe

ChrisRisner:
Adding first draft of the Java Code Reviews recipe including:

- Style guide
- Code analysis
- Code formatting
- Build validation
- Code review checklist


code-reviews/recipes/Java.md
```diff
@@ -0,0 +1,30 @@
+# Java Code Reviews
+
+## Java Style Guide
+
+[CSE](../../CSE.md) developers generally follow the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html).
+
+## Code Analysis
```
TessFerrandez: Should we change to Linting to follow the other recipes?

code-reviews/recipes/Java.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Build Validation
+
+It's important to enforce your code style and rules in the CI to avoid any team members merging code that does not comply with standards into your git repo.  If building using Azure DevOps, the [Maven and Gradle build tasks support](https://docs.microsoft.com/en-us/azure/devops/java/standalone-tools?view=azure-devops) using [PMD](https://pmd.github.io/), [Checkstyle](https://checkstyle.sourceforge.io/), and [FindBugs](http://findbugs.sourceforge.net/) code analysis tools as part of every build.
+
+## Code Review Checklist
+
+* [ ] Does the project use Lambda to make code cleaner?
+* [ ] Is dependency injection (DI) used?  Is it setup correctly?
+* [ ] If you're using Spring Boot, are you using @Inject instead of @Autowire?
```
TessFerrandez: Nit: If you're using -> If the code uses Spring Boot 

since this is a code review check point

code-reviews/recipes/Java.md
```diff
@@ -0,0 +1,30 @@
...
+Also frequently used is [FindBugs](http://findbugs.sourceforge.net/) and [PMD](https://pmd.github.io/).
+
+## Automatic Code Formatting
+
+Eclipse, and other Java IDEs, support automatic code formatting.  If using Maven, some developers also make use of the [formatter-maven-plugin](https://github.com/revelc/formatter-maven-plugin).
+
+## Build Validation
+
+It's important to enforce your code style and rules in the CI to avoid any team members merging code that does not comply with standards into your git repo.  If building using Azure DevOps, the [Maven and Gradle build tasks support](https://docs.microsoft.com/en-us/azure/devops/java/standalone-tools?view=azure-devops) using [PMD](https://pmd.github.io/), [Checkstyle](https://checkstyle.sourceforge.io/), and [FindBugs](http://findbugs.sourceforge.net/) code analysis tools as part of every build.
+
```
TessFerrandez: Can we add a small example of a .yml file section with the steps?

code-reviews/recipes/Java.md
debugerr: I think we are only referencing analyzers, so some of the recipes would be better off with Code Analysis, but I think the definition of Linter vs. Analyzers are very close anyway and used interchangeably when I look through the web. Maybe we can use "Linting / Analyzing" as a compromise and update the other sections? @TessFerrandez

ChrisRisner: I'm fine with that approach as I do think there is a lot of overlap.  If @TessFerrandez agrees, I can update that here and then create an issue and update the other docs as well.

code-reviews/recipes/Java.md
ChrisRisner: Added, please review latest commit changes.

code-reviews/recipes/Java.md
ChrisRisner: Addressed in latest commit.

code-reviews/recipes/Java.md
```diff
@@ -0,0 +1,53 @@
...
+    - task: Maven@3
+    displayName: 'Maven pom.xml'
+    inputs:
+        mavenPomFile: '$(Parameters.mavenPOMFile)'
+        checkStyleRunAnalysis: true
+        pmdRunAnalysis: true
+        findBugsRunAnalysis: true
+```
+
+Here is an example yaml for a Gradle build task with all three analysis tools enabled:
```
debugerr: Thanks Chris. Looks good to me.

code-reviews/recipes/Java.md
debugerr: @TessFerrandez, I would like to close this out today. If you don't agree with the suggestion pls chime in.

TessFerrandez: sounds good to me - we had this discussion on the C# recipe and IIRC you (or someone else) wanted to change to Linters...  as long as it's consistent I am good with either

PR #248, opened 2020-03-19
Proposing new fundamental "Developer Experience"

charleszipp:
This pull request proposes the addition of a new fundamental "Developer Experience". This something the dallas and cambridge crews focused heavily in our previous project. Midway through our project we invested heavily on making the "essential tasks" as easy as possible. Through that effort we saw significant reduction in regressions as well as velocity.

The essential spirit of this is the more we invest in making these essential tasks easy, the more time the team has to focus on value-add activities (coding and testing the feature being implemented).

cc @xtophs @SaraSp @DanMass (will also add Tracey once i can find her github id)

### DIFF
--- a/developer-experience/readme.md
+++ b/developer-experience/readme.md
@@ -1,0 +1,130 @@
+# Developer Experience
+
+Developer experience refers to the how easy or difficult it is for a developer to perform essential tasks needed to implement a change. A positive developer experience would mean these tasks are relatively easy for the team (see measures below).
+
+The essential tasks are identified below.
+
+Build - Verify that changes are free of syntax error and compile.
+Test - Verify that all automated tests pass.
+Start - Launch end-to-end to simulate execution in a deployed environment.
+Debug - Attach debugger to started solution, set breakpoints, step through code, and inspect variables.
+
+If effort is invested to make these activities as easy as possible, the returns on that effort will be multiply the longer the project runs and the larger the team.
+
+## Goals
+
+- Maximize the amount of time engineers spend on writing code that fulfills story acceptance and done-done criteria.
+- Minimize the amount of time spent manual setup and configuration of tooling
+- Minimize regressions and new defects by making end-to-end testing easy
+
+## Impact
+
+Developer experience can have a significant impact on the efficiency of the day to day execution of the team. A positive experience can pay dividends throughout the lifetime of the project; especially as new developers join the team.
+
+- Increased Velocity - Team spends less time on non-value-add activities such as dev/local environment setup, waiting on remote environments to test, and rework (fixing defects).
+- Improved Quality - When its easy to debug and test, developers will do more of it. This will translate to fewer defects being introduced.
+- Easier Onboarding & Adoption - When dev essential tasks are automated, there is less documentation to write and, subsequently, less to read to get started!
+
+## Measures
+
+### Time to First Debug (aka F5 Contract)
+
+Assuming a laptop/pc that has never run the solution, how long does it take to start debugging the solution end-to-end.
+
+### Time To First Commit
+
+How long does it take to make a change that can be verified/tested locally. A locally verified/tested change is one that passes test cases without introducing regression or breaking changes.
+
+## Participation
+
+The entire development team should take responsibility in ensuring positive
+
+## Facilitation Guidance
+
+The following outline several strategies that can be adopted to promote a positive developer experience.
+
+### Establish Hotkeys
+
+Assign hotkeys to each of the essential tasks.
+
+[`TODO: Suggest hotkeys for mac`]
+
+| Task                 | Windows      |
+| -------------------- | ------------ |
+| Build                | CTRL+SHIFT+B |
+| Test                 | CTRL+R,T     |
+| Start With Debugging | F5           |
+
+### The F5 Contract
+
+The F5 contract aims for the ability to run the end-to-end solution with the following steps.
+
+1. Clone - git clone <my-repo>
+2. Configure - set any configuration values that need to be unique to the individual (i.e. update a .env file)
+3. Press F5 - launch the solution with debugging attached.
+
+Most IDE's have some form of a task runner that can be used to automate the build, execute, and attach steps. Try to leverage these such that the steps can all be run with as few manual steps as possible.
+
+### Pay Attention to the Struggles
+
+Ask the team to surface the areas they spend significant time in that they consider non-value-add activities. For example, do they need to deploy their changes to an environment off their laptop before they can validate if what they did worked. Rather than debugging locally, do they have to do this repetitively to get to a working solution? Does this take several minutes each iteration? Does this block other developers due to the contention on the environment?
+
+The following are a few strategies for surfacing dev experience struggles
+
+- Retrospectives. Create a card that describes the struggle
+- Standup Blockers. Call out being blocked not being able to test change and why that cant be done.
+
+#### Fixing the Struggles
+
+Once raised, use the following steps to capture and create **actionable** fix.
+
+1. Define the problem - End to end testing is only possible when the solution is deployed to environment in Azure
+2. Identify a potential solution - Ability to run automated end-to-end tests on the developer machine.
+3. Justify fixing the problem - The team has identified this issue as the root cause for 3 stories not completing within their sprint over the past 2 sprints.
+
+### Single IDE Instance
+
+Avoid requiring multiple instances of the IDE to running concurrently to run and debug the solution locally. This will negatively impact the above measure. Additionally, that negative impact is multiplied by each instance required.
+
+- Configuration steps * IDE instances
+- Build steps * IDE instances
+- Start/Debug steps * IDE instances
+- Stop steps * IDE instances
+- Run test steps * IDE instances
+- Documenting all of the above * IDE instances
+
+### Single Repository
+
+Splitting a solution across multiple repositories can negatively impact the above measures. This can also negatively impact other areas such as Pull Requests, Automated Testing, Continuous Integration, and Continuous Delivery. Similar to the IDE instances, the negative impact is multiplied by the number of repositories.
+
+- clone steps * # of repos
+- branching steps * # of repos
+- commit steps * # of repos
+- CI * # of repos
+- Pull request reviews & merges * # of repos
+
+#### Atomic Pull Requests
+
+When the solution is encapsulated within a single repository, it also allows pull requests to represent a change across multiple layers. This is especially helpful when a change requires changes to a shared contract between multiple components. For example, a story requires that an api endpoint is changed. With this strategy the api and web client could be updated with the same pull request. This avoids master being broken temporarily while waiting on dependent pull requests to merge.
+
+[`TODO: Add link to other documented benefits of consolidated repos within source control section`]
+
+### When to Split Into Multiple Repositories
+
+[`TODO: Add link to source control section if it has guidance on this subject`]
+
+(Typically try to have the repository align with the product backlog.)
+
+### Minimize Remote Dependencies for Local Development
+
+The fewer dependencies on components that cannot run a developer's machine translate to fewer steps required to get started. Therefore, fewer dependencies will positively impact the measures above.
+
+The following strategies can be used to reduce these dependencies
+
+#### Use an Emulator
+
+If available, emulators are implementations of technologies that are typically only available in cloud environments. A good example is the [CosmosDB emulator](https://docs.microsoft.com/en-us/azure/cosmos-db/local-emulator).
+
+#### Feature Toggle a Mock or Different Technology
+
+Abstract the layer that has the remote dependency behind an interface owned by the solution (not the remote dependency). Create an implementation of that interface using a technology that can be run locally. Lastly, incorporate a feature flag that can be set to control which implementation is used.
--- a/developer-experience/readme.md
+++ b/developer-experience/readme.md
@@ -11,6 +11,10 @@ The essential tasks are identified below.
 
 If effort is invested to make these activities as easy as possible, **the returns on that effort will be multiply the longer the project runs and the larger the team**.
 
+## Defining End-to-End
+
+This document makes several references to running a solution end-to-end (aka E2E). End-to-end for the purposes of this document is scoped to the software that is owned, built, and shipped by the team. Systems owned by other teams or third-party vendors is not within the E2E scope for the purposes of this document.
+
 ## Goals
 
 - Maximize the amount of time engineers spend on writing code that fulfills story acceptance and done-done criteria.
@@ -87,7 +91,7 @@ Avoid requiring multiple instances of the IDE to running concurrently to run and
 - Run test steps * IDE instances
 - Documenting all of the above * IDE instances
 
-### Single Repository
+### Minimize the Number of Repositories
 
 Splitting a solution across multiple repositories can negatively impact the above measures. This can also negatively impact other areas such as Pull Requests, Automated Testing, Continuous Integration, and Continuous Delivery. Similar to the IDE instances, the negative impact is multiplied by the number of repositories.
 
--- a/developer-experience/readme.md
+++ b/developer-experience/readme.md
@@ -59,7 +59,7 @@ Assign hotkeys to each of the essential tasks.
 
 The F5 contract aims for the ability to run the end-to-end solution with the following steps.
 
-1. Clone - git clone <my-repo>
+1. Clone - git clone [`my-repo-url-here`]
 2. Configure - set any configuration values that need to be unique to the individual (i.e. update a .env file)
 3. Press F5 - launch the solution with debugging attached.
 
--- a/developer-experience/readme.md
+++ b/developer-experience/readme.md
@@ -49,10 +49,12 @@ Providing a positive developer experience is a team effort. However, certain mem
 
 The following are examples of how the Tech Lead might set the bar for dev experience
 
-- Determines development environment (IDE, hosting, etc)
+- Determines development environment (suggested IDE, hosting, etc)
 - Determines source control environment and number of repo's required
 - Given development environment and repo structure sets expectations for team to meet in terms of steps to perform the essential dev tasks
 - Nominates the DevEx Champion
+
+IDE choice is NOT intended to mandate that all team members must use the same IDE. However, this choice will direct where tight-integration investment will be prioritized. For example, if Visual Studio Code is the **suggested** IDE then, the team would focus on integrating VS code tasks and launch configurations over similar integrations for other IDE's. Team members should still feel free to use their preferred IDE as long as it does not negatively impact the team.
 
 ### DevEx Champion - Identify Iterative Improvements
 
@@ -60,12 +62,12 @@ The devex champion takes ownership in holding the team accountable for providing
 
 - Actively seek opportunities for improving the solution developer experience
 - Work with the tech lead to iteratively improve team expectations for developer experience
-- Curate a backlog actionable stories that identify areas for improvement.
+- Curate a backlog actionable stories that identify areas for improvement and prioritize with respect to project delivery goals by engaging directly with the Product Owner and Customer.
 - Serve as subject matter expert for the rest of the team. Help the team determine how to implement devex expectations and identify deviations.
 
-### Senior Members - Assert Expectations
+### Team Members - Assert Expectations
 
-The senior members of the team can also help hold the team accountable for providing a positive developer experience. The following are examples of areas the senior members can help identify where the team's devex expectations are not being met.
+The team members of the team can also help hold each other accountable for providing a positive developer experience. The following are examples of areas team members can help identify where the team's devex expectations are not being met.
 
 - Pull requests. Try the changes locally to see if they are adhering to the team's devex expectations.
 - Design Reviews. Look for proposals that may negatively effect the solution's devex. These might include
--- a/developer-experience/readme.md
+++ b/developer-experience/readme.md
@@ -43,7 +43,23 @@ How long does it take to make a change that can be verified/tested locally. A lo
 
 ## Participation
 
-The entire development team should take responsibility in ensuring positive experience. How
+### Tech Lead - Set the bar
+
+The following are examples of how the Tech Lead might set the bar for dev experience
+
+- Determines development environment (IDE, hosting, etc)
+- Determines source control environment and number of repo's required
+- Given development environment and repo structure sets expectations for team to meet in terms of steps to perform the essential dev tasks
+
+### Senior Members - Assert Expectations
+
+The following are examples of how the senior engineers on the team can assert the expectation set by the lead.
+
+[`TODO: Add bullet list of examples`]
+
+### DevEx Champion
+
+[`TODO: Nominate an individual on the team to be the dev ex chamption`]
 
 ## Facilitation Guidance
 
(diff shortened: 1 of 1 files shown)


developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Goals
+
+- Maximize the amount of time engineers spend on writing code that fulfills story acceptance and done-done criteria.
+- Minimize the amount of time spent manual setup and configuration of tooling
+- Minimize regressions and new defects by making end-to-end testing easy
+
+## Impact
+
+Developer experience can have a significant impact on the efficiency of the day to day execution of the team. A positive experience can pay dividends throughout the lifetime of the project; especially as new developers join the team.
```
jplane: Dev productivity and reduced friction during the actual CSE engagement period (while very important) are arguably the least interesting aspect... its the accrual of those benefits long after we disengage from active work with the customer which is our big goal, I think. We should call that out here, explicitly.

developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+
+- Increased Velocity - Team spends less time on non-value-add activities such as dev/local environment setup, waiting on remote environments to test, and rework (fixing defects).
+- Improved Quality - When its easy to debug and test, developers will do more of it. This will translate to fewer defects being introduced.
+- Easier Onboarding & Adoption - When dev essential tasks are automated, there is less documentation to write and, subsequently, less to read to get started!
+
+## Measures
+
+### Time to First Debug (aka F5 Contract)
```
jplane: I like this... but I wonder if its better to focus on "time to first E2E result"? Certainly getting F5 to work is important, but a focus on "time to setup and run the whole system E2E and see a result" assumes other best practices are in place, like setting up the simplest E2E scenario as early as possible, etc.

developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+
+Assuming a laptop/pc that has never run the solution, how long does it take to start debugging the solution end-to-end.
+
+### Time To First Commit
```
jplane: I feel like "time to first F5" is subsumed by this. So in line with my last comment... I think we should focus on "time to execute the system without debugging" and "time to first commit... which by definition encompasses other important milestones in the dev cycle like finding/reading docs, pulling a story, debugging, PR review, etc."

developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+
+How long does it take to make a change that can be verified/tested locally. A locally verified/tested change is one that passes test cases without introducing regression or breaking changes.
+
+## Participation
+
+The entire development team should take responsibility in ensuring positive
```
jplane: Yes... but let's recommend an ongoing owner for this. Smells like a team lead responsibility. Could also be delegate formally to another senior team member.

developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Facilitation Guidance
```
jplane: Specific examples here are fine... but let's leave room (and a clear mandate/expectation) for each team/project to define their own devex and (importantly) refine it over time via standard feedback mechanisms (retros etc.)

developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+
+### The F5 Contract
+
+The F5 contract aims for the ability to run the end-to-end solution with the following steps.
+
+1. Clone - git clone [`my-repo-url-here`]
+2. Configure - set any configuration values that need to be unique to the individual (i.e. update a .env file)
+3. Press F5 - launch the solution with debugging attached.
+
+Most IDE's have some form of a task runner that can be used to automate the build, execute, and attach steps. Try to leverage these such that the steps can all be run with as few manual steps as possible.
```
jplane: Might be good to point to (or include here) some example VS Code task files, etc.

developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Pay Attention to the Struggles
+
+Ask the team to surface the areas they spend significant time in that they consider non-value-add activities. For example, do they need to deploy their changes to an environment off their laptop before they can validate if what they did worked. Rather than debugging locally, do they have to do this repetitively to get to a working solution? Does this take several minutes each iteration? Does this block other developers due to the contention on the environment?
```
jplane: Doubling down on the ownership aspect... the owner of project devex should proactively seek out feedback on how the current approach is working, where it can improve, soliciting best practices from other teams, and then incorporating stories/tasks for improvement items into sprint planning.

Retros are of course a way to circulate this information, but we shouldn't merely wait for problems to bubble up. Let's promote a mindset of seeking them out and eliminating them, actively.

developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+- Retrospectives. Create a card that describes the struggle
+- Standup Blockers. Call out being blocked not being able to test change and why that cant be done.
+
+#### Fixing the Struggles
+
+Once raised, use the following steps to capture and create **actionable** fix.
+
+1. Define the problem - End to end testing is only possible when the solution is deployed to environment in Azure
+2. Identify a potential solution - Ability to run automated end-to-end tests on the developer machine.
+3. Justify fixing the problem - The team has identified this issue as the root cause for 3 stories not completing within their sprint over the past 2 sprints.
```
jplane: I think this criteria can vary on a per-project basis... not sure we need (want?) to be quite so prescriptive here.

developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Single IDE Instance
```
jplane: This one seems overly prescriptive to me... ?

developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+
+Splitting a solution across multiple repositories can negatively impact the above measures. This can also negatively impact other areas such as Pull Requests, Automated Testing, Continuous Integration, and Continuous Delivery. Similar to the IDE instances, the negative impact is multiplied by the number of repositories.
+
+- clone steps * # of repos
+- branching steps * # of repos
+- commit steps * # of repos
+- CI * # of repos
+- Pull request reviews & merges * # of repos
+
+#### Atomic Pull Requests
```
jplane: Totally agree... and I think this could be pulled up under the "Single Repo" item... ?

developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+
+The fewer dependencies on components that cannot run a developer's machine translate to fewer steps required to get started. Therefore, fewer dependencies will positively impact the measures above.
+
+The following strategies can be used to reduce these dependencies
+
+#### Use an Emulator
+
+If available, emulators are implementations of technologies that are typically only available in cloud environments. A good example is the [CosmosDB emulator](https://docs.microsoft.com/en-us/azure/cosmos-db/local-emulator).
+
+#### Feature Toggle a Mock or Different Technology
```
jplane: I guess depending on scope of the abstracted element a feature toggle might be appropriate... for more narrow abstractions DI can be better and is worth a mention here.

developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+
+- Increased Velocity - Team spends less time on non-value-add activities such as dev/local environment setup, waiting on remote environments to test, and rework (fixing defects).
+- Improved Quality - When its easy to debug and test, developers will do more of it. This will translate to fewer defects being introduced.
+- Easier Onboarding & Adoption - When dev essential tasks are automated, there is less documentation to write and, subsequently, less to read to get started!
+
+## Measures
+
+### Time to First Debug (aka F5 Contract)
+
+Assuming a laptop/pc that has never run the solution, how long does it take to start debugging the solution end-to-end.
```
nmiodice: Nit: "end to end" is a bit of an overloaded terms for projects that fit into (or will fit into) a larger ecosystem of software. I think in this case you intend "end to end" to mean running the software that is being implemented as part of the engagement, and it might be worth it to rephrase to indicate that, or define what "end to end" means here.

developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+
+The entire development team should take responsibility in ensuring positive
+
+## Facilitation Guidance
+
+The following outline several strategies that can be adopted to promote a positive developer experience.
+
+### Establish Hotkeys
+
+Assign hotkeys to each of the essential tasks.
```
nmiodice: As long as we are proposing this section as *examples of strategies*, and not *strategies that must be employed*, I think this is OK. We should avoid making an assumption that people will be using the same IDE, which is why this is a good *example* of a strategy.

developer-experience/readme.md
nmiodice: Not sure what the guidance would be for somebody who prefers to develop purely in a terminal, but is it possible to generalize this section to focus on the core problem being discussed here without assuming a specific technology?

Here are some thoughts on how this may be more generally written:
 - Standardizing the configuration for solution components
 - Standardizing the way tests are run for each component
 - Standardizing on a small number of technology stacks
 - Standardizing on documentation styles and practices

developer-experience/readme.md
```diff
@@ -0,0 +1,130 @@
...
+Avoid requiring multiple instances of the IDE to running concurrently to run and debug the solution locally. This will negatively impact the above measure. Additionally, that negative impact is multiplied by each instance required.
+
+- Configuration steps * IDE instances
+- Build steps * IDE instances
+- Start/Debug steps * IDE instances
+- Stop steps * IDE instances
+- Run test steps * IDE instances
+- Documenting all of the above * IDE instances
+
+### Single Repository
```
nmiodice: I think the real problem is having a lot of repositories *without a way to effectively manage build artifacts and versioned dependencies.* 

I like the phrasing you have in the description below, as it focuses on keeping a single *solution* in one repository. That's key, but it does not match your title for this section. Perhaps a re-wording could be *Minimizing number of repositories*

developer-experience/readme.md
charleszipp: Adding bolded text that calls this out
**Most importantly, the customer will continue to accrue these benefits long after the code-with engagement.**

developer-experience/readme.md
charleszipp: Rewording title and description to adjust focus

developer-experience/readme.md
charleszipp: If im understanding the feedback right, this might be resolved by the adjustment to Time to First Debug. Lets revisit this one once i push my change for the earlier comment.

developer-experience/readme.md
charleszipp: Rewording the top description to this... @nmiodice i believe had similar feedback

> The following outline examples of several strategies that can be adopted to promote a positive developer experience. It is expected that each team should define what a positive dev experience means within the context of their project. Additionally, refine that over time via feedback mechanisms such as sprint and project retrospectives.

developer-experience/readme.md
charleszipp: Yeah after re-reading this in questioning how valuable this section is. Think i will replace this with something that clarifies ownership aspect

developer-experience/readme.md
charleszipp: removing

developer-experience/readme.md
charleszipp: It is a subheader under the Single Repo header. Am I misunderstanding the suggestion?

developer-experience/readme.md
charleszipp: That was what i was trying to convey but, probably not doing it successfully. 

In my experience, you would use DI to inject a different implementation of a shared interface. However, that is also usually encapsulated within some type of factory logic where a decision is made based on some env config/setting. I was referring to that practice as feature toggling but, perhaps that leaves to much to interpretation (or is the wrong interpretation?). 

Maybe need to be more explicit? Thoughts?

developer-experience/readme.md
charleszipp: Good point. Agree this leaves much room for interpretation. I added a section at the start to define what it means in the context of this doc..

> ## Defining End-to-End
> 
> This document makes several references to running a solution end-to-end (aka E2E). End-to-end for the purposes of this document is scoped to the software that is owned, built, and shipped by the team. Systems owned by other teams or third-party vendors is not within the E2E scope for the purposes of this document.

developer-experience/readme.md
charleszipp: I reworded this based on @jplane feedback to be more suggestive rather than mandatory. See if that wording aligns better with what you are suggesting.

developer-experience/readme.md
charleszipp: I think this is closely related to the single repo section so maybe it should be a subsection within there. However, the pain i have felt from past projects that have a lots of repo's is you end up having to manage multiple instances of VSCode, Visual Studio, Sublime Text or whatever your preferred IDE is. Also this almost always multiplies the number of steps it takes to configure, build, test, and start the entire solution. 

This issue also leads teams to start to favor horizontal sliced work as a means to mitigate all the inefficiencies with managing all the repo's and IDE's. This starts to create even more issues down the road.

So i think what im trying to suggest is that being to develop in a single IDE is a strategy to prevent this from happening. Does that clarification change what your suggestion?

developer-experience/readme.md
charleszipp: Thinking about this more @jplane, i put this rough outline together. What are your thoughts? The DevEx Champion represents your suggestion about formal delegation to an individual. cc @nmiodice 

### Tech Lead - Set the bar

The following are examples of how the Tech Lead might set the bar for dev experience

- Determines development environment (IDE, hosting, etc)
- Determines source control environment and number of repo's required
- Given development environment and repo structure sets expectations for team to meet in terms of steps to perform the essential dev tasks

### Senior Members - Assert Expectations

The following are examples of how the senior engineers on the team can assert the expectation set by the lead.

[`TODO: Add bullet list of examples`]

### Developer Experience Champion

[`TODO: Nominate an individual on the team to be the dev ex chamption`]

developer-experience/readme.md
charleszipp: Great point, will reword this to assign ownership of this aspect to the DevEx Champion

developer-experience/readme.md
jplane: Love it.

One additional suggestion... add another bullet under Tech Lead for "nominates/selects project DevEx champion... this could be the lead him/herself, or could be another senior team member"

jplane: Also, per my comment below... let's outline a few basic responsibilities for the champion. I'm personally less interested in a cookie-cutter recipe... more interested in emphasizing iterative, proactive responsibility for care and feeding of devex for the duration of the project.

developer-experience/readme.md
jplane: I hear you... when I saw "feature toggling" I read "feature flags" which is conceptually related to but distinct from DI (different scopes, different mechanisms, different rules of engagement, etc.). I'd recommend the word be a bit more explicit.

developer-experience/readme.md
charleszipp: Updated section. Please take a look and see if there are any other changes you think should be made. 

I agree of the emphasis. However, I would like to root this in a concrete practice to which the reader can relate. This can give some guidance on where to start then adapt as needed. Its not my intent to provide a recipe. More so to root in concrete examples. Hopefully this change reflects that. Let me know if you think it does not.

developer-experience/readme.md
charleszipp: Updated and provided example. let me know if this better represents what you were suggesting.

developer-experience/readme.md
jplane: my mistake, i'm an idiot :)

developer-experience/readme.md
nmiodice: I think that the "multiple IDE" problem you are describing is the solution to the underlying issue that you described well in your comment:
> Also this almost always multiplies the number of steps it takes to configure, build, test, and start the entire solution. ... This issue also leads teams to start to favor horizontal sliced work as a means to mitigate all the inefficiencies with managing all the repo's and IDE's. This starts to create even more issues down the road.

My fear is that somebody takes the "single IDE" suggestion literally but does not fix the underlying issue. An example of this is my current project, where I use 1 IDE instance to open 10 git repositories at any given time. The challenge is still there, even though I only need 1 IDE instance for my work.

developer-experience/readme.md
```diff
@@ -0,0 +1,203 @@
+# Developer Experience
+
+Developer experience refers to the how easy or difficult it is for a developer to perform essential tasks needed to implement a change. A positive developer experience would mean these tasks are relatively easy for the team (see measures below).
+
+The essential tasks are identified below.
```
nmiodice: Not sure if this is in scope here or not, but do you see things like "finding the backlog" or "understanding the board" as falling into this proposed fundamental?

developer-experience/readme.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Participation
+
+Providing a positive developer experience is a team effort. However, certain members can take ownership of different areas to help hold the entire team accountable.
+
+### Tech Lead - Set the bar
+
+The following are examples of how the Tech Lead might set the bar for dev experience
+
+- Determines development environment (IDE, hosting, etc)
```
nmiodice: IDE seems like something that developers should be free to choose, but I do see value in sharing common configurations for sensible choices.

I.e., a checkstyle configuration for Intellij can be shared across team members without mandating that all developers should use that IDE

developer-experience/readme.md
```diff
@@ -0,0 +1,203 @@
...
+- Given development environment and repo structure sets expectations for team to meet in terms of steps to perform the essential dev tasks
+- Nominates the DevEx Champion
+
+### DevEx Champion - Identify Iterative Improvements
+
+The devex champion takes ownership in holding the team accountable for providing a positive developer experience. The following outline responsibilities for the DevEx champion.
+
+- Actively seek opportunities for improving the solution developer experience
+- Work with the tech lead to iteratively improve team expectations for developer experience
+- Curate a backlog actionable stories that identify areas for improvement.
```
nmiodice:
```suggestion
- Curate a backlog actionable stories that identify areas for improvement and prioritize with respect to project delivery goals by engaging directly with the Product Owner and Customer

```

developer-experience/readme.md
```diff
@@ -0,0 +1,203 @@
...
+- Serve as subject matter expert for the rest of the team. Help the team determine how to implement devex expectations and identify deviations.
+
+### Senior Members - Assert Expectations
```
nmiodice: While I imagine that somebody with a "senior" title may do this more proactively, I think that non "senior" members hold a very valuable perspective here as well.

developer-experience/readme.md
```diff
@@ -0,0 +1,203 @@
...
+
+- clone steps * # of repos
+- branching steps * # of repos
+- commit steps * # of repos
+- CI * # of repos
+- Pull request reviews & merges * # of repos
+
+#### Atomic Pull Requests
+
+When the solution is encapsulated within a single repository, it also allows pull requests to represent a change across multiple layers. This is especially helpful when a change requires changes to a shared contract between multiple components. For example, a story requires that an api endpoint is changed. With this strategy the api and web client could be updated with the same pull request. This avoids master being broken temporarily while waiting on dependent pull requests to merge.
```
nmiodice: Nice

developer-experience/readme.md
```diff
@@ -0,0 +1,203 @@
...
+
+The fewer dependencies on components that cannot run a developer's machine translate to fewer steps required to get started. Therefore, fewer dependencies will positively impact the measures above.
+
+The following strategies can be used to reduce these dependencies
+
+#### Use an Emulator
+
+If available, emulators are implementations of technologies that are typically only available in cloud environments. A good example is the [CosmosDB emulator](https://docs.microsoft.com/en-us/azure/cosmos-db/local-emulator).
+
+#### Use DI + Toggle to Mock Remote Dependencies
```
nmiodice: This is a pretty deep topic, is there any guidance we can link to instead of adding code samples here?

developer-experience/readme.md
charleszipp: There is a section in the playbook for backlog management that i believe would be responsible for what you are calling out here. Agree though it definitely can create churn if work is hard to find or ambiguous.

developer-experience/readme.md
charleszipp: I think i can probably reword this to suggest that the team invest in deep integration with a single IDE. Other engineers are free to use something else but, may not be able to gain the benefits of the deep integration set with the default.

developer-experience/readme.md
charleszipp: Agree, i think we got the ownership aspect @jplane was after with the first two headings (tech lead and devex champion)

developer-experience/readme.md
charleszipp: One of the many benefits when we consolidated repos on my last project ;)

developer-experience/readme.md
charleszipp: @nmiodice I replaced this section with two others that combine your very valid point about standardization of essential across components and automating those tasks across the solution. I think this takes an approach at describing a potential solution that is not necessarily tied to an IDE. For example, I think that someone could easily see implementing a solution level build with a shell script would be fine (in the case they are hard-core enough to be using notepad/sublime + terminal). Let me know what you think of this update and if it better focuses on the root problem/solution.

developer-experience/readme.md
charleszipp: Adding this note to highlight this is a suggestion, not a mandate

> IDE choice is NOT intended to mandate that all team members must use a single IDE. However, this choice will direct where tight-integration investment will be prioritized. For example, if Visual Studio Code is the **suggested** IDE then, the team would focus on integrating VS code tasks and launch configurations over similar integrations for other IDE's. Team members should still feel free to use their preferred IDE as long as it does not negatively impact the team.

developer-experience/readme.md
charleszipp: Adding suggestion...

developer-experience/readme.md
charleszipp: changing to team members rather than senior...

charleszipp:
> I like where this is going... how about something like this to further illustrate the point ?
> 
> [Configuration steps +Build steps + Start/Debug steps + Stop steps + Run test steps + Documenting all of the above] * [many solution components] = TOO MANY STEPS
> 
> VS.
> 
> [Configuration steps +Build steps + Start/Debug steps + Stop steps + Run test steps + Documenting all of the above] * [1 solution] = MINIMUM NUMBER OF STEPS

I think that makes it much more clear @jplane making that change. Think we should do similar format for the repos?

developer-experience/readme.md
charleszipp: I did a quick search to see if i could find any existing write-ups for this specific use case but was not able to. Since the code sample and supporting guidance is already written could we allow the first version to keep this? I think if we find ourselves adjusting this section in the future we should probably spend more effort finding a relevant article or, write our own that we can link to. Looks like there might be a sharing opportunity here since i wasnt able to initially find what i was looking for.

TessFerrandez:
Is there a corresponding issue?

charleszipp:
> Is there a corresponding issue? if so, can you link it?

Hey @TessFerrandez, there is no corresponding issue for this one. This didnt come out of one of the ef champions groups. It was added as sharing activities between projects.

PR #252, opened 2020-03-21
Adding initial testing file structure

gled4er:
Hello,

This PR contains a proposal for automated testing file structure.

@RichCzyzewski  I am looking forward for your feedback.

This PR is addressing the corresponding issue - https://github.com/microsoft/code-with-engineering-playbook/issues/275

Thank you!


TessFerrandez:
@gled4er Can you add a link to the issue this PR addresses so that it is closed at the end of the PR, and also a bit more context in the PR about what is being addressed so that we can review against that criteria?

automated-testing/ab-testing/readme.md
```diff
@@ -0,0 +1,23 @@
+# ~Test type~
```
TessFerrandez: Is there a reason for chosing strike through characters in the title ~Test type~ and later on in the document, it looks very odd - assuming this is because it should be replaced at some point with AB-testing in this case, why not write AB-testing at once?

automated-testing/ab-testing/readme.md
```diff
@@ -0,0 +1,23 @@
...
+
+The document should start with a brief overview about the test type and what is covered in this document, the goal here is to provide a high-level description to help the reader understand what is covered to decide whether to continue reading or not.
```
TessFerrandez: This is a question to debate, but should we add documents as TODO:s? It is not completely apparent that this is a TODO until you start reading more thoroughly.  

IMHO at any point in time **master** is "shippable" so an alternative way to fix this is by instead creating issues for the items that need to be filled in, with templates in the issues, and either leave empty documents - perhaps saying "Information on this topic will be added soon" - or maybe not even adding links right now, but noting in the tasks that links, and docs need to be added.

continuous-integration/CICD.md
```diff
@@ -2,7 +2,7 @@
 
 Continuous Integration is the engineering practice of frequently committing code in a shared repository, ideally several times a day, and performing an automated build on it. These changes are built with other simultaneous changes to the system, which enables early detection of integration issues between multiple developers working on a project. Build breaks due to integration failures are treated as the highest priority issue for all of the developers on a team and generally work stops until they are fixed.
 
-Paired with a test driven development approach, continuous integration also allows us to also test the integrated build such that we can verify that not only does the code base still build correctly, but also is still functionally correct. This is also a best practice for building robust and flexible software systems.
+Paired with automated testing approach, continuous integration also allows us to also test the integrated build such that we can verify that not only does the code base still build correctly, but also is still functionally correct. This is also a best practice for building robust and flexible software systems.
```
icchan: Paired with **an** automated testing approach...

automated-testing/readme.md
```diff
@@ -0,0 +1,16 @@
+# Testing
+
+## Sections within Testing
+
+* [Unit testing](unit-testing/readme.md)
+* [Unit testing v2](unit-testing-v2/readme.md)
```
fnocera: What would be the difference between unit testing v2 and unit testing? Seems strange to me to have two sections talking about the same area differently.

automated-testing/ab-testing/readme.md
alkofahi: This was used as a placeholder to be changed later, I agree that it does not look good but it is meant to be changed. We can take out the strikethrough and replace it with something else?

automated-testing/readme.md
```diff
@@ -0,0 +1,16 @@
...
+* [Integration testing](integration-testing/readme.md)
```
fnocera: This is a LOT of different types of testing, getting content for all of these would definitely be our end goal, however I am wondering whether the focus should be on getting a few of these (most used) done first vs putting placeholders/templates for each of these.

alkofahi: +1 - I don't think we should check in the template as content for each readme. Building this incrementally makes sense as well vs putting placeholders.

gled4er:
Hello @TessFerrandez ,

I added the issue - https://github.com/microsoft/code-with-engineering-playbook/issues/275

Thank you!

automated-testing/ab-testing/readme.md
gled4er: Hello @TessFerrandez ,

The idea of this PR was to create placeholder pages with the template we picked up so for the contributors that will finish each of the pages will be straightforward to follow the same structure.

Thank you!

continuous-integration/CICD.md
gled4er: Fixed. Thank you @icchan !

automated-testing/readme.md
gled4er: Hello @fnocera ,

We did not want to modify the existing Unit Testing version before the new one is created. That is why we created new page for the new version we will be working on. 

Thank you!

automated-testing/readme.md
gled4er: Thank you for the feedback. The idea of the whole PR was to create placeholder pages based on the template we picked. I am ok to modify the approach. Let's sync today and decide on our strategy. 

Thank you!

automated-testing/ab-testing/readme.md
TessFerrandez: I saw that you started a conversation on teams about this, and I think it's fair if the other changes will come in immediately.  Just wanted to bring it up as this goes against our own best practices:

https://github.com/microsoft/code-with-engineering-playbook/blob/master/source-control/readme.md
- Commit complete and well tested code. Never commit incomplete code, get in the habit of testing your code before committing.

automated-testing/readme.md
TessFerrandez: I agree with @fnocera here, leaving the docs in an "in-between" state is a bit risky as many people are using these docs.

gled4er:
Closing this PR and moving to writing about each type of test.

PR #281, opened 2020-04-10
Add security rules of engagement

anthturner:
Provides insight into what standard rules testers must follow as well as how an application must be prepped.


doymturner:
Are details provided anywhere on how to request the security review? I'm not seeing them. 

Also, should the rules of engagement be expanded to include what should be provided during the request with regards to planning attacks (e.g. IPs, URLs)? This could especially important when testing an application deployed within a customer's infrastructure/account.

anthturner:
We're still fleshing out how requesting security reviews will look like behind the scenes. I can wait to merge this until we're done with that, but I'm not clear on how we'll break apart public guidance from internal process (given the audience of this material).

doymturner:
@anthturner it was more of a question of making sure it's accounted for in the plan somewhere due to it being mentioned in this PR. Sounds like it's in the planning pipeline which is good enough to cover my question.

anthturner:
FWIW -- the request process is much more concrete now (SME request). I'll be submitting a separate PR to cover more information about engaging.

fnocera:
Added @doymturner to the contributors so now this can be merged in @anthturner

PR #295, opened 2020-05-06
Redisign the Q&A section

### LINKED ISSUE #259: Redesign Q&A section
Currently the Q&A section on the main README is only around some specific areas. 

I would suggest we create a tips & tricks section with problem statements and tips on how to solve these issues. I imagine a collection of problems we ran into in our engagement and some tricks on how to solve them or improve the process.

e.g. Our code reviews are too slow! We are progressing not fast enough
- nominate a add a code review champion that keeps an eye on slow moving PRs
- use ADO tools (custom styles for PRs that are not progressing) to support the process
- reserve 5min in the standup to go through pending PRs that need review

All of the above are scattered around in our content. It would be nice to create a condensed version of tips people can turn to, if they run into a problem with the CR process.

debugerr:
- move the Q&A out of the main readme and add a new FAQ file
- add common issues / questions to the file and links to our content

closes #259

### DIFF
--- a/code-reviews/faq.md
+++ b/code-reviews/faq.md
@@ -1,0 +1,37 @@
+# FAQ
+
+This is a list of questions / frequently occurring issues when working with code reviews and answers how you can possibly tackle them.
+
+## We experience very large PRs, how can we fix this?
+
+Make sure you size the work items smaller. They should be able to be completed on its own. The team is instructed to commit early, before the full product backlog item / user story is complete, but rather when an individual item is done. If the work would result in an incomplete feature, make sure it can be turned off, until the full feature was delivered.
+More information can be found in [Pull Requests - Size Guidance](./pull-requests.md#size-guidance).
+
+## We experience slow code reviews, causing delays in delivering features
+
+### Possible actions you can take:
+
+- Add a rule for PR turnaround time to your work agreement.
+- Setup a slot after the standup to go through pending PRs and assign the ones that are inactive
+- Dedicate a PR review master who will be responsible to keep things flowing by assigning or notifying people when PR got stale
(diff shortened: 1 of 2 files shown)


code-reviews/faq.md
```diff
@@ -0,0 +1,37 @@
+# FAQ
+
+This is a list of questions / frequently occurring issues when working with code reviews and answers how you can possibly tackle them.
+
+## We experience very large PRs, how can we fix this?
+
+Make sure you size the work items smaller. They should be able to be completed on its own. The team is instructed to commit early, before the full product backlog item / user story is complete, but rather when an individual item is done. If the work would result in an incomplete feature, make sure it can be turned off, until the full feature was delivered.
```
TessFerrandez, line 7: its own => their own

ChrisRisner: also "was delivered" => "is delivered"

code-reviews/faq.md
```diff
@@ -0,0 +1,37 @@
...
+More information can be found in [Pull Requests - Size Guidance](./pull-requests.md#size-guidance).
+
+## We experience slow code reviews, causing delays in delivering features
+
+### Possible actions you can take:
+
+- Add a rule for PR turnaround time to your work agreement.
+- Setup a slot after the standup to go through pending PRs and assign the ones that are inactive
```
ChrisRisner, line 15: This and the following line seem to be the outliers on using periods at the end of bullet items or not with both of these not having them but others all having them.  Would recommend adding here.

code-reviews/faq.md
debugerr: fixed

code-reviews/faq.md
debugerr: done.

PR #309, opened 2020-05-22
Add more examples of "Ready" + Add docs for creating issue templates

nmiodice:
This change adds more examples of ready checklist items to the
definition of "Ready." It also adds links to documentation that can
help to codify this checklist as issue templates.


team-agreements/definition-of-ready/readme.md
```diff
@@ -16,6 +16,8 @@ It can be understood as a checklist that helps the Product Owner to ensure that
 * [ ] Does the user story have clear and complete acceptance criteria?
 * [ ] Does the user story address the business need?
 * [ ] Can we measure the acceptance criteria?
+* [ ] Can the story be broken down into smaller stories?
```
charleszipp, line 19: Would suggest changing to more concrete target such as. This will help communicate when a story is small enough.

> Can the story be completed within a single sprint? 

This shouldnt be taken as law (i.e. no story that spans a sprint can be accepted by the team). However, it should be the exception and very intentional decision by the team to take a story that they know will span multiple sprints.

team-agreements/definition-of-ready/readme.md
```diff
@@ -2,11 +2,11 @@
...
+> If a user story is not Ready in the beginning of the Sprint it increases the chance that the story will not be done at the end of this sprint.
 
 ## What it is
 
-Definition of Ready is the agreement made by the scrum team around how complete a user story should be in order to be selected as candidate for estimation in the sprint planning.
+*Definition of Ready* is the agreement made by the scrum team around how complete a user story should be in order to be selected as candidate for estimation in the sprint planning. These can be codified as a checklist in user stories using [Github Issue Templates](https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository) or [Azure DevOps Work Item Templates](https://docs.microsoft.com/en-us/azure/devops/boards/backlogs/work-item-template?view=azure-devops&tabs=browser).
 
 It can be understood as a checklist that helps the Product Owner to ensure that the user story he wrote contains all the necessary details for the scrum team to understand the work to be done.
```
charleszipp, line 11: I know you didnt write this but the pronoun use should be fixed

> ...to ensure that the user story **they** wrote contains...

team-agreements/definition-of-ready/readme.md
```diff
@@ -16,6 +16,8 @@ It can be understood as a checklist that helps the Product Owner to ensure that
 * [ ] Does the user story have clear and complete acceptance criteria?
 * [ ] Does the user story address the business need?
 * [ ] Can we measure the acceptance criteria?
+* [ ] Can the story be broken down into smaller stories?
+* [ ] Is the story blocked?
```
charleszipp, line 20: I think it would be worth providing some examples of blocked scenarios. This mostly comes up with dependencies/predecessors. For example, this story cant be completed until this other story is finished. Another common case is we are waiting on some sort of decision external to the team before we can complete (i.e. theres an outstanding question that needs to be answered by LT or another team.)

nmiodice:
Thanks @charleszipp - I just went through and updated some of the changes and added a section on how to make work ready. LMK your thoughts!

team-agreements/definition-of-ready/readme.md
```diff
@@ -2,20 +2,24 @@
...
-It can be understood as a checklist that helps the Product Owner to ensure that the user story he wrote contains all the necessary details for the scrum team to understand the work to be done.
+It can be understood as a checklist that helps the Product Owner to ensure that the user story they wrote contains all the necessary details for the scrum team to understand the work to be done.
 
 ### Examples of ready checklist items:
 
 * [ ] Does the description have the details including any input values required to implement the user story?
 * [ ] Does the user story have clear and complete acceptance criteria?
 * [ ] Does the user story address the business need?
 * [ ] Can we measure the acceptance criteria?
+* [ ] Can the story be completed within a single sprint? If not, can it be broken down into smaller stories?
```
madossan01, line 19: Despite the fact that verifying if a user story can be completed within a single sprint is a good practice, it wouldn't be a blocker to the user store be considered ready. The sprint planning will be the right place to define this as the feasibility of completing the user story will depend on the team capacity in the sprint. So, I suggest to change by something like this: "Is the user story accomplishing a scope that can not be split into meaningful parts that will attend the customer needs?"

team-agreements/definition-of-ready/readme.md
```diff
@@ -2,20 +2,24 @@
...
+* [ ] Is the story blocked? For example, does it depend on any of the following:
```
madossan01, line 20: These are important examples to include in the checklist, thank you. Only a minor reminder to replace "story" by "user story" to keep consistency across the document.

team-agreements/definition-of-ready/readme.md
nmiodice: I'm trying to think through this and am having a tough time coming up with the correct phrasing. The suggested phrasing:

> Is the user story accomplishing a scope that can not be split into meaningful parts that will attend the customer needs?

Feels very similar to the item a few lines above:

> Does the user story address the business need?

Do you agree? What I am trying to indicate here is that the story should aim to be a minimal slice of functionality that should be completed as part of the same "unit". What are your thoughts on this phrasing, borrowing some phrases from [here](https://martinfowler.com/bliki/UserStory.html)?

> Is the user story small enough to be implemented in a short amount of time, but large enough to provide value to the customer?

madossan01: Good point and I found myself thinking for a while on how to better phrase this question in my first review. The rephrasing you proposed: "Is the user story small enough to be implemented ..." is much better than mine, sure we can keep it. Thanks.

nmiodice:
Thanks for your review @madossan01 - I believe I've addressed each of your comments - let me know if I've missed anything!

PR #324, opened 2020-06-26
Suggesting XUnit instead of NUnit

dariuszparys:
I would suggest to declare XUnit instead of NUnit, most engagements I've worked on used that and also it is used by some PG's as well.


omusavi:
Hey Dariusz, we actually are planning on making some big changes to this page in particular, and one of those changes will be to avoid making recommendations between many of these testing frameworks.

In the case of xUnit vs NUnit, both are very well supported by .NET Core, and much of the tooling supporting unit testing, including through Visual Studio. While there are advantages to both, I think ultimately it falls down to personal preference to choose between them. In cases where you could easily go either way I don't think we should choose a single framework, instead pitching both and maybe a short comparison where appropriate. 

I hope to have the new unit testing page up for review very soon, so I encourage you to come check it out and leave feedback once it has been revised!

PR #331, opened 2020-07-03
Replace scrum master with process lead

### LINKED ISSUE #328: Rename scrum master to process lead
I have a local branch to fix this just need to be added to Collaborators team.
The "Process Lead" role consists of both scrum and agile process management.

tessmichi:
# Replace scrum master with process lead

## Link to the design document that prompted this PR

- Relevant Teams threads can be found in CSE General as AMA question responses as well as dedicated post from Michael Brown.

- "Scrum Master" should not be the term we use for multiple reasons. Firstly, the connotation of the usage of the word "master." Secondly, this Process Lead should lead the team through processes outside of scrum, including agile and other processes the team agrees to use.

## Description of new changes

- This is just a documentation change
- I found all uses of "scrum master" and replaced with "process lead"
- In places where needed, I explained briefly what a process lead is.

## How did you test it

- N/A

## Any relevant logs or outputs

- N/A

### Other information or known dependencies

- Fixes #328

### DIFF
--- a/sprint-planning/readme.md
+++ b/sprint-planning/readme.md
@@ -22,7 +22,7 @@ General guidance:
 
 Specific roles:
 
-- [Scrum Master](https://www.agilealliance.org/glossary/scrum-master):
+- [Process Lead](https://www.agilealliance.org/glossary/scrum-master):
   - Facilitate the conversation.
   - Ensure everyone is heard.
   - Remind scrums principles and sprint planning goals if necessary.
--- a/sprint-planning/readme.md
+++ b/sprint-planning/readme.md
@@ -22,10 +22,10 @@ General guidance:
 
 Specific roles:
 
-- [Process Lead](https://www.agilealliance.org/glossary/scrum-master):
+- [Process Lead]:
   - Facilitate the conversation.
   - Ensure everyone is heard.
-  - Remind scrums principles and sprint planning goals if necessary.
+  - Remind scrums/agile/other principles and sprint planning goals if necessary, updating the working agreement where needed to ensure a mapping between principals and what is working/not working for the team.
 - [Product owner](https://www.agilealliance.org/glossary/product-owner/):
   - Prior to the sprint planning: performs some [backlog grooming](../backlog-management/grooming/readme.md) to ensure that each story that they want to propose for the new sprint (*) :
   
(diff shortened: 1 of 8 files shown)


tessmichi:
Thanks for taking a look Tess!

Process Lead was actually coined by @wbreza and our team has been using it for a few months now.
However, outside our team, there was a thread posted in CSE FTE - General (I won't link it as the link does contain tenant information that isn't private but I don't think belongs in a public repo anyway) last week where folks discuss names.
There is definitely not total consensus in any direction, but I believe there to be enough support of this term that we can use it to get rid of the historical "scrum master". There has already been collaboration on the new Process Lead page in the cwcwiki, and folks who had originally proposed different titles are hopping on on with Process Lead now. Process Lead seems to be the path of least resistance here, and easier for more folks to get behind if it wasn't their first choice.

Mostly, the key difference is that having "scrum" in the name suggests that responsibilities will map to the "scrum master certification" which they won't, as we'd like this person to also handle agile processes etc. Additionally, keeping "scrum" out of the name allows teams to discuss on their engagement kickoff what they want to fall under the roles and responsibilities, rather than treating the recommendation as a mandate.
In regards to "coach" instead of "lead", my decision here was mainly based on the multiple differing interpretations of the rugby analogy. Some think it's Captain, some think Coach. I worry that picking one will be too confusing to others.

tessmichi:
Here is the cwcwiki page link: https://cwcwiki.com/wiki/Process_Lead
I had been thinking of adding the list here as well, but it felt redundant. Should I add it here too? And if so, is the working agreement the best place?

tessmichi:
Given some comments on the thread asking for it in the fundamentals repo, I have added recommended responsibilities in the working agreement.

sprint-planning/readme.md
```diff
@@ -22,7 +22,7 @@ General guidance:
 
 Specific roles:
 
-- [Scrum Master](https://www.agilealliance.org/glossary/scrum-master):
+- [Process Lead](https://www.agilealliance.org/glossary/scrum-master):
```
bsherwin, line 25: Does it still make sense to link out to Agile Alliance here or does that cause confusion?

tessmichi: That's a great point, removing and will resolve when gone

tessmichi: Pushed, and added a little more wordiness to the bullet point referencing this bit. Do you want to review and close this issue if it meets your goal?

PR #353, opened 2020-07-28
Github Actions Environment Variables at runtime

KanishkT123:
This PR includes the following:

1. Guidance on adding non secret runtime variables into commit messages for Github Actions
2. Guidance on adding non secret runtime variables into the PR body for Github Actions

Both of these make CI/CD pipeline authoring and testing significantly easier, as they allow for certain stages to be skipped, tags to be added to outputs, or for flags to be manually set. This can significantly increase inner loop dev time, as it means that integration pipelines do not need to constantly make external calls during pipeline testing, long CI/CD processes that are not directly relevant for debugging can be skipped, etc.


KanishkT123:
I would suggest the CI/CD label. 

I also just realized that the branch name convention is incorrect, I do apologize for that.

github-actions/runtime-variables/RuntimeVariables.MD
```diff
@@ -0,0 +1,234 @@
+# Runtime Variables in Github Actions
+
+## Objective
+While Github Actions is a popular choice for writing and running CI/CD pipelines, especially for open source projects hosted on Github, it lacks specific quality of life features found in other CI/CD environments. One key feature that Github Actions has not yet implemented is the ability to **mock and inject runtime variables** into a workflow, in order to test the pipeline itself.
```
gfmatthews, line 4: I would make the comparisson here that we are attempting to bridge a feature that exists in DevOps to a feature that doesn't exist in GitHub in a conceptually similar way.

github-actions/runtime-variables/RuntimeVariables.MD
```diff
@@ -0,0 +1,234 @@
...
+2. This triggers the workflow (as will any push). As the `[commit var]` is **not** in the commit message, the `${COMMIT_VAR}` variable in the workflow will be set to `false` and result in the following:
+
+![Commit True Scenario](images/CommitFalse.PNG)
+
+
+### PR Body Variables
+
+When a PR is made, the PR Body can also be used to setup variables. These variables can be made available to all the workflow runs that stem from that PR, which can help ensure that commit messages are more informative and less cluttered, and reduces the work on the developer.
+
+Once again, this for an expected key and value. In this case, the key is `PR_VAR` and the value is `[pr var]`.
```
gfmatthews, line 163: It might be worth noting that with PR templates, you can "systematize" this idea a bit more across a team

github-actions/runtime-variables/RuntimeVariables.MD
```diff
@@ -0,0 +1,234 @@
...
+```
+- name: "Set flag from PR"
+  env:
+    PR_VAR: ${{ contains(github.event.pull_request.body, '[pr var]') }}
+  run: |
+  echo "::set-env name=flag::${PR_VAR}"
+  echo "set flag to ${PR_VAR}"
+```
+
+#### Usage
```
gfmatthews, line 226: I might consider adding a section for real-world usage similar to how we used this concept in a previous project. What other things could developers consider doing with this? I.e. creating on-demand integration environments, setting runtime configs, etc

github-actions/runtime-variables/RuntimeVariables.MD
```diff
@@ -0,0 +1,234 @@
+# Runtime Variables in Github Actions
```
charleszipp, line 1: This content would probably be more appropriate to be under continous-deployment/recipes/github-actions-runtime-variables

KanishkT123: Done! I've changed the specific folder location after talking to Erik Schlegel.

github-actions/runtime-variables/RuntimeVariables.MD
KanishkT123: Done!

github-actions/runtime-variables/RuntimeVariables.MD
KanishkT123: Done! I did this in the objective after adding an `in scope` and `out of scope` section, with the out of scope serving as a place for "ideas that someone else should do but I don't think belong in this guide".

github-actions/runtime-variables/RuntimeVariables.MD
KanishkT123: Done!

PR #354, opened 2020-07-28
Add guidance for continuous deployment environments and GitHub workflows

manojvazirani:
This PR includes the following:
1. Basic overview of continuous deployment
1. Guidance on creating effective pipelines for CI/CD
1. Basic introduction to Github workflows
1. Guidance on creating workflows for different environments


continuous-deployment/pipelines/recipes/github-workflows/workflow-per-environment.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Workflow Per Environment
+
+The general approach is to have one pipeline, where the code is built, tested and deployed and the artifact is then promoted to the next environment, eventually to be deployed into production.
+
+There are multiple ways in GitHub that an environment setup can be achieved. The way it can be done is to have one workflow for multiple environments, which does not mean it cannot be done, but the complexity increases as additional processes and jobs are added to a workflow. One way to get around the complexity is to have separate workflows for different environments, making sure that the artifacts from each workflow are independent, as well as, the workflow is small enough to debug any issues seen in any of the workflows.
+
+This also helps to keep the deployments to the environments independent thus reducing the time to deploy and find issues earlier than later in the process. Also sine the environments are independent of each other, any failures in deploying to one environment does not block deployments to other environments.
+
```
juzuluag, line 24: @manojvazirani , one small comment here. Maybe something to keep in mind is the state of the environment when you have  'serialized' pipelines.

continuous-deployment/pipelines/recipes/github-workflows/workflow-per-environment.md
```diff
@@ -0,0 +1,31 @@
...
+## Workflow Per Environment
+
+The general approach is to have one pipeline, where the code is built, tested and deployed and the artifact is then promoted to the next environment, eventually to be deployed into production.
+
+There are multiple ways in GitHub that an environment setup can be achieved. The way it can be done is to have one workflow for multiple environments, which does not mean it cannot be done, but the complexity increases as additional processes and jobs are added to a workflow. One way to get around the complexity is to have separate workflows for different environments, making sure that the artifacts from each workflow are independent, as well as, the workflow is small enough to debug any issues seen in any of the workflows.
+
+This also helps to keep the deployments to the environments independent thus reducing the time to deploy and find issues earlier than later in the process. Also sine the environments are independent of each other, any failures in deploying to one environment does not block deployments to other environments.
```
juzuluag, line 23: @manojvazirani , I think there is a typo ("sine")

continuous-deployment/pipelines/readme.md
```diff
@@ -0,0 +1,17 @@
...
+
+Pipelines is a way to automatically build and test your code project and make it available to other users. It works with just about any language or project type.
+
+Pipelines combines continuous integration (CI) and continuous delivery (CD) to constantly and consistently test and build your code and ship it to any target.
+
+A few things to keep in mind when creating pipelines for continuous deployment:
+
+1. Keep your pipelines fast
+1. Isolate and secure your CI/CD environment
+1. Make the CI/CD pipeline the only way to deploy to production
```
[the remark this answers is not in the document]
manojvazirani: Thats an awesome catch. I updated the header for the list on line 7 to suggestions, since all the above points are suggestions.

continuous-deployment/pipelines/recipes/github-workflows/workflow-per-environment.md
manojvazirani: Thats an interesting point. I added a few comments for state of the environments in both of the approaches.

continuous-deployment/readme.md
```diff
@@ -1,7 +1,14 @@
 # Continuous Deployment
```
charleszipp, line 1: I would suggest trying to consolidate the content you put under `Pipelines` and here. Seems that the recipe is really trying to describe how to implement Continuous **Delivery** or **Deployment** (yes there is a distinct difference between these) across multiple environments using GitHub Actions. Pipelines term really only applies in AzDO and Jenkins AFAIK. So readers might find this confusing. The readers will more associate your content as a solution to CD rather than pipeline.

continuous-deployment/pipelines/recipes/github-workflows/workflow-per-environment.md
```diff
@@ -0,0 +1,31 @@
+# GitHub Workflows
+
+A workflow is a configurable automated process made up of one or more jobs where each of these jobs can be an action in GitHub. Currently a YAML file format is supported for defining a workflow in GitHub.
+
+Additional information on GitHub actions and GitHub Workflows in the links posted in the [references](#References) section below.
+
+## Environments
```
charleszipp, line 7: Seems like this should be in the root CD readme

continuous-deployment/readme.md
```diff
@@ -1,7 +1,14 @@
...
+Continuous deployments can be achieved by creating different environments in [Azure DevOps](https://docs.microsoft.com/en-us/aspnet/core/azure/devops/cicd) or by using multiple [GitHub workflows](https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow). These environments can be used for different reasons like some environments are used for development and testing while others are used for production.
+
+More details on environments and the steps for setting up different environments using GitHub workflows can be found [here](../continuous-deployment/pipelines/recipes/github-workflows/workflow-per-environment.md).
+
+For additional information and details on CI/CD please refer to this [link](../continuous-integration/CICD.md).
 
 ## Sections within Continuous Deployment
 
 * [Secrets Management](secrets-management/readme.md)
+* [Pipelines](pipelines/readme.md)
```
charleszipp, line 14: I might add a link to the explanation of difference between Continuous Delivery and Deployment. Also, make sure the content is consistent on which one its describing. Example article here... https://www.atlassian.com/continuous-delivery/principles/continuous-integration-vs-delivery-vs-deployment

Delivery = Ship on time bound intervals. Also commonly referred to as Release Trains in larger dev shops
Deployment = Ship on commit all the way to prod.

Both are automated but, delivery will require someone to click a button to ship. Deployment lands in prod with absolutely no human intervention.

continuous-deployment/pipelines/recipes/github-workflows/workflow-per-environment.md
```diff
@@ -0,0 +1,31 @@
...
+
+One way to get around the complexity of a single workflow is to have separate workflows for different environments, making sure that the artifacts from each workflow are independent, as well as, the workflow is small enough to debug any issues seen in any of the workflows. In this case, the state and environment values need to be passed from one deployment environment to another. Multiple workflows also helps to keep the deployments to the environments independent thus reducing the time to deploy and find issues earlier than later in the process. Also, since the environments are independent of each other, any failures in deploying to one environment does not block deployments to other environments.
```
charleszipp, line 23: > ...making sure that the artifacts from each workflow are independent...

Im not sure I agree with this statement. Are you suggesting that binaries and/or container images be rebuilt within each workflow? How are you going to ensure you have the exact same code as you promote to new env's? I think the only way you could do that without potentially getting other commits mixed in is to target commit hash in each workflow.

charleszipp, line 23: A picture would go a long way to better convey what is being explained here. Perhaps a sequence or flowchart? This would help people that are more visual like myself.

continuous-deployment/pipelines/readme.md
```diff
@@ -0,0 +1,17 @@
+# Pipelines
+
+Pipelines is a way to automatically build and test your code project and make it available to other users. It works with just about any language or project type.
+
+Pipelines combines continuous integration (CI) and continuous delivery (CD) to constantly and consistently test and build your code and ship it to any target.
+
+There are a few suggestions listed below to keep in mind when creating pipelines for continuous deployment:
+
+1. Keep your pipelines fast
```
gfmatthews, line 9: Nit: "As fast as feasible". Ideally I'd run a full test battery on every single build, but we all know we are conscious about trade-offs when making dev builds or smoke builds.

continuous-deployment/pipelines/readme.md
```diff
@@ -0,0 +1,17 @@
...
+1. Isolate and secure your CI/CD environment
+1. Make the CI/CD pipeline the only way to deploy to production
+1. Maintain parity with production wherever possible
+1. Build only once and promote the result through the pipeline
+1. Run your fastest tests early
```
gfmatthews, line 14: So that you can fail out of the build pipeline fast and free up resources for others?

continuous-deployment/pipelines/recipes/github-workflows/workflow-per-environment.md
```diff
@@ -0,0 +1,31 @@
...
+
+To understand more about deployment environments, and their differences, please refer to the link in the [references](#References) section below.
+
+## Workflow Per Environment
+
+The general approach is to have one pipeline, where the code is built, tested and deployed and the artifact is then promoted to the next environment, eventually to be deployed into production.
+
+There are multiple ways in GitHub that an environment setup can be achieved. The way it can be done is to have one workflow for multiple environments, but the complexity increases as additional processes and jobs are added to a workflow, which does not mean it cannot be done for small pipelines. The plus point of having one workflow is that, when an artifact flows from one environment to another the state and environment values between the deployment environments can be passed easily.
```
gfmatthews, line 21: "One way" it can be done :)

continuous-deployment/pipelines/recipes/github-workflows/workflow-per-environment.md
gfmatthews: I do worry a little that recommending the creation of separate but very much closely related workflows might become a maintenance issue down the line too.

continuous-deployment/readme.md
manojvazirani: Thats an interesting view on things @charleszipp. Will consolidate and update the two areas.

continuous-deployment/pipelines/recipes/github-workflows/workflow-per-environment.md
manojvazirani: @gfmatthews : Thats totally true and will add it as a tradeoff for having separate pipelines.

continuous-deployment/pipelines/readme.md
manojvazirani: Yes, as well as unblock other pipelines to run if limited resources are available. Will update statement to reflect that.

continuous-deployment/pipelines/recipes/github-workflows/workflow-per-environment.md
charleszipp: As we build more pipelines on github actions, would like to revisit this with more samples. However, think its good starting point for now. 
I am still not sure I agree with whats being said here but, understand available feature set in github actions might be influencing the choices made.

manojvazirani: I see you make an interesting point, and as we see github actions and workflows is evolving everyday, there will be changes coming in the future for better flow of artifacts.

PR #359, opened 2020-08-05
Added Load Testing Section

icchan:
Created the load test document as a supplement to the performance test section.

It only focusses on load test, and I tried not to overlap with performance test too much.


automated-testing/performance-testing/load-testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+Additionally the results of a load test can also be used as data to help with capacity planning and calculating scalability. 
+
+## Load Testing Design Blocks
+
+There are a number of basic component which are required to carry out a load test.  
+
+1. In order to have meaningful results the system needs to be tested in a production-like environment with a network and hardware which closely resembles the expected deployment environment.
+
+2. The load will consist consist a module which simulates user activity. Of course what the composition of this "user activity" will be vary based on the type of application being tested. For example an e-commerce web site might simulate user browsing and purchasing items, but an IoT data ingestion pipeline would simulate a stream of device readings.
```
h2floh, line 22:
```suggestion
2. The load will consist of a module which simulates user activity. Of course what the composition of this "user activity" will be vary based on the type of application being tested. For example an e-commerce web site might simulate user browsing and purchasing items, but an IoT data ingestion pipeline would simulate a stream of device readings.

```

automated-testing/performance-testing/load-testing.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Planning
+
+1. **Identify key scenarios to measure**  - Gather these scenarios from Product Owner, they should provide a representative sample of real world traffic.
+2. **Determine expected normal and peak load for the scenarios** -  Determine a load level such as concurrent users or requests per second to find the size of the load test you will run.
+3. **Identify success criteria metrics** - These may be testing side such as response time and error rate, or it may be on the system side such as CPU and memory usage.
+
+### Execution
+
+It is recommended to use an existing testing framework (see below). These tools will provide a method of both specifying the user activity scenarios as well as how to execute those at load. It is common to slowly ramp up to your desired load just to be safe. Once you have reached your defined workload.
```
maniSbindra, line 36: Can we consider modifying "just to be safe". One of the reasons of Ramp up and ramp down is to mimic the real world application

automated-testing/performance-testing/load-testing.md
```diff
@@ -0,0 +1,69 @@
...
+It is recommended to use an existing testing framework (see below). These tools will provide a method of both specifying the user activity scenarios as well as how to execute those at load. It is common to slowly ramp up to your desired load just to be safe. Once you have reached your defined workload. 
+
+Before starting please be aware of any restrictions on your network such as DDOS protection where you may need to notify a network administrator or apply for an exemption.
+
+### Further Testing
+After completing your load test you should be set up to continue on to additional related testing such as;
+- **Soak Testing** - Performing a load test over an extended period of time to ensure long term stability.
+- **Stress Testing** - Gradually increasing the load to find the limits of the system and identify the maximum capacity.
+- **Spike Testing** - Introduce a sharp short-term increase into the load scenarios.
+- **Scalability Testing** - Re-testing of a system as your expand horizontally or vertically to measure how it scales.
```
maniSbindra, line 47: Should the descriptions be added under the performance testing page, and then linked from here?

automated-testing/performance-testing/load-testing.md
```diff
@@ -0,0 +1,68 @@
...
+
+- **JMeter** (<https://jmeter.apache.org/>) - Has built in patterns to test without coding, but can be extended with Java.
+- **Artillery** (<https://artillery.io/>) - Write your scenarios in Javascript, executes a node application.
+- **Gatling** (<https://gatling.io/>) -  Write your scenarios in Scala with their DSL.
+- **Locust** (<https://locust.io/>) - Write your scenarios in Python using the concept of concurrent user activity.
+- **K6** (<https://k6.io/>) - Write your test scenarios in Javascript, available as open source or as SaaS.
+
+## Conclusion
+
+A load test is critical step to understand if a target system will be reliable under the expect real world traffic.
```
maniSbindra, line 61:
```suggestion
A load test is critical step to understand if a target system will be reliable under the expected real world traffic.

```

automated-testing/performance-testing/load-testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Load Testing Frameworks and Tools
+
+Here are a few popular load testing frameworks you may consider and the languages used to define your scenarios.
+
+- **JMeter** (<https://jmeter.apache.org/>) - Has built in patterns to test without coding, but can be extended with Java.
+- **Artillery** (<https://artillery.io/>) - Write your scenarios in Javascript, executes a node application.
+- **Gatling** (<https://gatling.io/>) -  Write your scenarios in Scala with their DSL.
+- **Locust** (<https://locust.io/>) - Write your scenarios in Python using the concept of concurrent user activity.
+- **K6** (<https://k6.io/>) - Write your test scenarios in Javascript, available as open source or as SaaS.
```
NT-D, line 57: nit: 
- I want to see this as table including name, language and description, so that I can understand easier.
- Want to know tools for C#, too

automated-testing/performance-testing/load-testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+A load test is critical step to understand if a target system will be reliable under the expected real world traffic.
+
+Of course its only as good as your ability to predict the expected load, so its important to follow up with other further testing to truly understand how your system behaves in different situations.
+
+## Resources
+
+List additional readings about this test type for those that would like to dive deeper.
+
+- [https://docs.microsoft.com/en-us/azure/architecture/framework/scalability/load-testing](https://docs.microsoft.com/en-us/azure/architecture/framework/scalability/load-testing)
```
NT-D, line 69: You may want to use `Load Testing`, page title, instead of url for easier understanding

automated-testing/performance-testing/load-testing.md
icchan: I didn't create a table because this is just a reference, not a comparison.
I added language because it seems to be the first thing people check when they decide to to write their scenarios

Do you know a good one for C#? The Azdo one has been deprecated so I didn't include it
https://devblogs.microsoft.com/devops/cloud-based-load-testing-service-eol/

NT-D: - I see, we don't need to add table here.
- Sorry I only know the Azdo one.. Let me update when I find good one.

automated-testing/performance-testing/load-testing.md
icchan: yeh fair point

automated-testing/performance-testing/load-testing.md
icchan: I thought about this, but i think it is meaningful here since they are specifically extensions of a load test 
however we can maybe reconsider this if we restructure or add additional sub pages.

automated-testing/performance-testing/load-testing.md
maniSbindra: @icchan is the plan to replace "just to be safe"?

automated-testing/performance-testing/load-testing.md
```diff
@@ -0,0 +1,69 @@
+
+# Load Testing
+
+"*Load testing is performed to determine a system's behavior under both normal and anticipated peak load conditions.*" - [Load testing - Wikipedia](https://en.wikipedia.org/wiki/Load_testing)
+
+A load test is designed to determine how a system behaves under expected normal workload.  Specifically its main purpose is to confirm if a system can handle the expected load level. Depending on the target system this could be concurrent users, requests per second or data size.
+
+## Why Load Testing
+
+The main objective prove the system can behave normally under the expected normal load before releasing it to production. The criteria which defines "behave normally" will depend on your target, this may be as simple as "the system remains available", but it could also include meeting a response time SLA or error rate.
```
RichCzyzewski, line 10: Nit: The main objective **is to** prove the

automated-testing/performance-testing/load-testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+Additionally the results of a load test can also be used as data to help with capacity planning and calculating scalability.
+
+## Load Testing Design Blocks
+
+There are a number of basic component which are required to carry out a load test.  
+
+1. In order to have meaningful results the system needs to be tested in a production-like environment with a network and hardware which closely resembles the expected deployment environment.
+
+2. The load will consist of a module which simulates user activity. Of course what the composition of this "user activity" will be vary based on the type of application being tested. For example an e-commerce web site might simulate user browsing and purchasing items, but an IoT data ingestion pipeline would simulate a stream of device readings.
```
RichCzyzewski, line 20: Perhaps we should have a note here that the preference is that the simulated data be as close to real data as possible in order to get correct results. for instance, if simulator data is too predictable caches could have cache/hit ratios far better than what production would look like.

RichCzyzewski, line 20: Nit: The load **test** will

automated-testing/performance-testing/load-testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Applying the Load Testing
+
+### Planning
+
+1. **Identify key scenarios to measure**  - Gather these scenarios from Product Owner, they should provide a representative sample of real world traffic.
+2. **Determine expected normal and peak load for the scenarios** -  Determine a load level such as concurrent users or requests per second to find the size of the load test you will run.
+3. **Identify success criteria metrics** - These may be on testing side such as response time and error rate, or they may be on the system side such as CPU and memory usage.
+
+### Execution
```
RichCzyzewski, line 34: Do we want to mention that it matters where the load test is run from and that this should be considered? For instance if testing a cloud component, running the load test agent from your local machine may give you inaccurate results for a handful of reasons (bandwidth, throttling, latency, etc).

automated-testing/performance-testing/load-testing.md
```diff
@@ -0,0 +1,69 @@
+
+# Load Testing
+
+"*Load testing is performed to determine a system's behavior under both normal and anticipated peak load conditions.*" - [Load testing - Wikipedia](https://en.wikipedia.org/wiki/Load_testing)
+
+A load test is designed to determine how a system behaves under expected normal workload.  Specifically its main purpose is to confirm if a system can handle the expected load level. Depending on the target system this could be concurrent users, requests per second or data size.
```
RichCzyzewski, line 6: Maybe mention both normal and peak in the first sentence.

automated-testing/performance-testing/load-testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+3. The load test will be initiated from a component external to the target system which can control the amount of load applied. This can be a single agent, but may need to scaled to multiple agents in order to achieve higher levels of activity.
+
+4. Although not required to run a load test, it is advisable to have monitoring and/or logging in place to be able to measure the impact of the test and discover potential bottlenecks.
+
+## Applying the Load Testing
+
+### Planning
+
+1. **Identify key scenarios to measure**  - Gather these scenarios from Product Owner, they should provide a representative sample of real world traffic.
```
RichCzyzewski, line 30: Part of planning could be also selected the proper load test tool. I've seen cases before where a load test tool was selected that could not generate the necessary traffic. So it might be good to indicate this is something to look at in selecting the proper tool.

maniSbindra (approved):
Lgtm

PR #414, opened 2020-10-19
Remove customer specifics

cteicher-m:
- customer specific information should not be in the playbook
- automated-testing/case-studies is empty.


fnocera (approved):

fnocera:
Wondering if we should remove the entirety of the case studies dir/folder?

cteicher-m:
> Wondering if we should remove the entirety of the case studies dir/folder?

I made that change locally and pushed. Perhaps that just isn't clear in the PR but will delete the whole folder?

PR #421, opened 2020-11-03
added bridge-to-kubernetes docs

plasne:
Added information about Bridge to Kubernetes after OneWeek testing.

### DIFF
--- a/developer-experience/recipes/bridge-to-kubernetes.md
+++ b/developer-experience/recipes/bridge-to-kubernetes.md
@@ -1,0 +1,60 @@
+
+# Bridge to Kubernetes
+
+Bridge to Kubernetes allows developers to seamlessly work on services both in a shared cluster and locally in debug mode.
+
+To learn more about Bridge to Kubernetes:
+
+- [Introduction](https://code.visualstudio.com/docs/containers/bridge-to-kubernetes)
+
+- [How it works](https://docs.microsoft.com/en-us/visualstudio/containers/overview-bridge-to-kubernetes)
+
+- [Configuration](https://docs.microsoft.com/en-us/visualstudio/containers/configure-bridge-to-kubernetes)
+
+- [VS Code Setup](https://docs.microsoft.com/en-us/visualstudio/containers/bridge-to-kubernetes?view=vs-2019)
+
+- [Repository](https://github.com/Microsoft/mindaro)
+
+## Useful Scenarios
+
+We believe this technology can be most helpful...
+
+- if a developer needs to test/debug a new or changed service against an ecosystem of services.
+
+- if a developer does not have the domain knowledge to deploy all the micro services. This allows them to work against a "blackbox" set of already deployed services.
+
+We found the technology very easy to implement so it can probably be used for a wide of cases with minimal training.
+
+## Limitations
+
+While many of these limitations are in the backlog, currently, you cannot use Bridge to Kubernetes in these circumstances...
+
+- if your services need to have replicas.
+
+- if your services require HTTPS and you are running in ISOLATION mode (ex. Istio mTLS).
+
+- if your Pods have more than 1 container (ex. sidecars, Istio).
+
+- if you are using WSL.
+
+- if you are using dev-containers.
+
+## Tested
+
+We have tested the following scenarios with and without ISOLATION mode:
+
+- from local service to shared service.
+
+- from shared service to local service.
+
+- from local service (computer A) to local service (computer B) - routing across AKS cluster.
+
+## Contact
+
+If you have questions, feel free to reach out to:
+
+- Peter Lasne (CSE)
+
+- Lila Molyva (CSE)
+
+- Product Group Alias: BridgeToKubernetes@microsoft.com
--- a/developer-experience/recipes/bridge-to-kubernetes.md
+++ b/developer-experience/recipes/bridge-to-kubernetes.md
@@ -6,47 +6,38 @@ Bridge to Kubernetes allows developers to seamlessly work on services both in a 
 To learn more about Bridge to Kubernetes:
 
 - [Introduction](https://code.visualstudio.com/docs/containers/bridge-to-kubernetes)
-
 - [How it works](https://docs.microsoft.com/en-us/visualstudio/containers/overview-bridge-to-kubernetes)
-
 - [Configuration](https://docs.microsoft.com/en-us/visualstudio/containers/configure-bridge-to-kubernetes)
-
 - [VS Code Setup](https://docs.microsoft.com/en-us/visualstudio/containers/bridge-to-kubernetes?view=vs-2019)
-
 - [Repository](https://github.com/Microsoft/mindaro)
 
 ## Useful Scenarios
 
-We believe this technology can be most helpful...
+This technology is most helpful...
 
 - if a developer needs to test/debug a new or changed service against an ecosystem of services.
-
 - if a developer does not have the domain knowledge to deploy all the micro services. This allows them to work against a "blackbox" set of already deployed services.
 
-We found the technology very easy to implement so it can probably be used for a wide of cases with minimal training.
+The technology can be used for a wide variety of cases with minimal training.
 
 ## Limitations
 
 While many of these limitations are in the backlog, currently, you cannot use Bridge to Kubernetes in these circumstances...
 
 - if your services need to have replicas.
-
 - if your services require HTTPS and you are running in ISOLATION mode (ex. Istio mTLS).
-
 - if your Pods have more than 1 container (ex. sidecars, Istio).
-
 - if you are using WSL.
-
 - if you are using dev-containers.
+
+Please check [here](https://docs.microsoft.com/en-us/visualstudio/containers/overview-bridge-to-kubernetes?view=vs-2019#limitations) for updated limitations.
 
 ## Tested
 
 We have tested the following scenarios with and without ISOLATION mode:
 
 - from local service to shared service.
-
 - from shared service to local service.
-
 - from local service (computer A) to local service (computer B) - routing across AKS cluster.
 
 ## Contact
@@ -54,7 +45,5 @@ We have tested the following scenarios with and without ISOLATION mode:
 If you have questions, feel free to reach out to:
 
 - Peter Lasne (CSE)
-
 - Lila Molyva (CSE)
-
 - Product Group Alias: BridgeToKubernetes@microsoft.com


developer-experience/recipes/bridge-to-kubernetes.md
```diff
@@ -0,0 +1,60 @@
+
+# Bridge to Kubernetes
+
+Bridge to Kubernetes allows developers to seamlessly work on services both in a shared cluster and locally in debug mode.
+
+To learn more about Bridge to Kubernetes:
+
+- [Introduction](https://code.visualstudio.com/docs/containers/bridge-to-kubernetes)
+
```
TessFerrandez, line 9: nit: remove empty lines in list

developer-experience/recipes/bridge-to-kubernetes.md
```diff
@@ -0,0 +1,60 @@
...
+- [VS Code Setup](https://docs.microsoft.com/en-us/visualstudio/containers/bridge-to-kubernetes?view=vs-2019)
+
+- [Repository](https://github.com/Microsoft/mindaro)
+
+## Useful Scenarios
+
+We believe this technology can be most helpful...
+
+- if a developer needs to test/debug a new or changed service against an ecosystem of services.
+
```
TessFerrandez, line 23: nit: remove empty lines in list

developer-experience/recipes/bridge-to-kubernetes.md
```diff
@@ -0,0 +1,60 @@
...
+- [Configuration](https://docs.microsoft.com/en-us/visualstudio/containers/configure-bridge-to-kubernetes)
+
+- [VS Code Setup](https://docs.microsoft.com/en-us/visualstudio/containers/bridge-to-kubernetes?view=vs-2019)
+
+- [Repository](https://github.com/Microsoft/mindaro)
+
+## Useful Scenarios
+
+We believe this technology can be most helpful...
```
TessFerrandez, line 20: nit: We believe this technology can be most helpful... => This technology is most helpful

developer-experience/recipes/bridge-to-kubernetes.md
```diff
@@ -0,0 +1,60 @@
...
+
+- if a developer needs to test/debug a new or changed service against an ecosystem of services.
+
+- if a developer does not have the domain knowledge to deploy all the micro services. This allows them to work against a "blackbox" set of already deployed services.
+
+We found the technology very easy to implement so it can probably be used for a wide of cases with minimal training.
```
TessFerrandez, line 26: nit: it looks a bit weird with personal reflection here - who is we? I would be more direct "The technology can be used for a wide variety of cases with minimal training"

also in the nit "wide of cases" => wide variety of cases or many different cases

developer-experience/recipes/bridge-to-kubernetes.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Limitations
+
+While many of these limitations are in the backlog, currently, you cannot use Bridge to Kubernetes in these circumstances...
+
+- if your services need to have replicas.
+
```
TessFerrandez, line 33: nit: remove the empty lines

developer-experience/recipes/bridge-to-kubernetes.md
```diff
@@ -0,0 +1,60 @@
...
+- if a developer needs to test/debug a new or changed service against an ecosystem of services.
+
+- if a developer does not have the domain knowledge to deploy all the micro services. This allows them to work against a "blackbox" set of already deployed services.
+
+We found the technology very easy to implement so it can probably be used for a wide of cases with minimal training.
+
+## Limitations
+
+While many of these limitations are in the backlog, currently, you cannot use Bridge to Kubernetes in these circumstances...
```
TessFerrandez, line 30: nit: remove the ...

developer-experience/recipes/bridge-to-kubernetes.md
```diff
@@ -0,0 +1,60 @@
...
+
+- if your Pods have more than 1 container (ex. sidecars, Istio).
+
+- if you are using WSL.
+
+- if you are using dev-containers.
+
+## Tested
+
+We have tested the following scenarios with and without ISOLATION mode:
```
TessFerrandez, line 44: nit: again - who are we? unless it is the authors of the kubernetes bridge (in which case we should state that), I think we should remove this section -- this section fits more in the bridge to kubernetes repo -- you don't want to have to update it both here and there when there are changes

developer-experience/recipes/bridge-to-kubernetes.md
TessFerrandez: Also, I would recommend removing this section all together and putting it in the bridge to kubernetes repo instead, otherwise you have to keep updating it in two places or this will become stale as soon as it is updated - Keep it DRY

developer-experience/recipes/bridge-to-kubernetes.md
```diff
@@ -0,0 +1,60 @@
...
+
+- from local service to shared service.
+
+- from shared service to local service.
+
+- from local service (computer A) to local service (computer B) - routing across AKS cluster.
+
+## Contact
+
+If you have questions, feel free to reach out to:
```
TessFerrandez, line 54: nit: we shouldn't add personal information in here - add this contact info to the bridge to kubernetes repo instead.  This is supposed to be read by many people outside of CSE with no connection to CSE - how would they contact a Peter Lasne in CSE? or know who that is

TessFerrandez (changes requested):

TessFerrandez:
Thank you for the contribution - I think it is great to have a reference to this, but a lot of this content fits better in that repo instead to avoid getting instantly out of sync.

developer-experience/recipes/bridge-to-kubernetes.md
plasne: done.

developer-experience/recipes/bridge-to-kubernetes.md
plasne: done.

developer-experience/recipes/bridge-to-kubernetes.md
plasne: I changed it as you suggested. This was written as part of a hackweek. So "we" was the team writing it.

developer-experience/recipes/bridge-to-kubernetes.md
plasne: done.

developer-experience/recipes/bridge-to-kubernetes.md
plasne: I added:
Please check [here](https://docs.microsoft.com/en-us/visualstudio/containers/overview-bridge-to-kubernetes?view=vs-2019#limitations) for updated limitations.

I don't maintain the bridge-to-kubernetes repo, but I still think think it is helpful to at least be aware there may be significant limitations and then they can follow the link and check.

developer-experience/recipes/bridge-to-kubernetes.md
plasne: done.

developer-experience/recipes/bridge-to-kubernetes.md
plasne: Frankly the only reason this document exists is because someone decided it would be valuable as a hack to validate that the bridge-to-kubernetes functionality worked as designed, was easy to use, and determine the revenant use-cases for CSE. Our team performed the hack and wrote the documentation in the code-with-engineering-playbook as discussed.

If someone outside CSE reads this and doesn't know who Peter or Lila is, then they cannot contact us and that is fine. But I was under the impression that generally this was for CSE and if they wanted more info on the hack they could contact us.

developer-experience/recipes/bridge-to-kubernetes.md
plasne: "We" is the CSE team that ran this investigation as a hack and the contacts. This hack was a validation of the technology put out by the team that wrote the kubernetes bridge. Their documentation states that they designed the solution to do lots of things. We tested these 3 things in the hack.

developer-experience/recipes/bridge-to-kubernetes.md
TessFerrandez: If that is the case - this content probably fits better in the CWC wiki - This playbook is supposed to be consumable by externals as well - both standalone and as leave behinds for partners we work with. From what you are mentioning here, this sounds like more of a CSE only resource - which fits better to the CWC wiki.

Thoughts @fnocera

fnocera: @plasne I tend to agree with @TessFerrandez that this might be better suited to the CWC wiki vs the playbook.
@charleszipp what are your thoughts on this given you are leading the developer experience portion of the playbook?

charleszipp: I agree about removing the contact information. As @TessFerrandez mentioned @plasne this content is intended to be consumable by the public. For that reason, I wouldnt put contact information as it would give the impression to the public that the two of you are supporting the feature. Additionally, I dont think we would want to give a direct email line of communication to the PG. 

I would suggest replacing this with something that directs people to post issues in the official github repo. There, the PG can triage issues as they are opened.

developer-experience/recipes/bridge-to-kubernetes.md
```diff
@@ -0,0 +1,49 @@
...
+
+To learn more about Bridge to Kubernetes:
+
+- [Introduction](https://code.visualstudio.com/docs/containers/bridge-to-kubernetes)
+- [How it works](https://docs.microsoft.com/en-us/visualstudio/containers/overview-bridge-to-kubernetes)
+- [Configuration](https://docs.microsoft.com/en-us/visualstudio/containers/configure-bridge-to-kubernetes)
+- [VS Code Setup](https://docs.microsoft.com/en-us/visualstudio/containers/bridge-to-kubernetes?view=vs-2019)
+- [Repository](https://github.com/Microsoft/mindaro)
+
+## Useful Scenarios
```
charleszipp, line 14: I would suggest also explaining how this relates with the content explained in the root readme. For example, what I suspect this relates to is helping with ease E2E start/debug; which is one of the essential tasks mentioned below. Also when would you use this approach. I believe you kind of hinted at this in the first line but, may be worth diving a bit deeper into when they might need to adopt this. 

For example, e2e debug requires integration with external services on the same cluster. Rather than run those services locally (which may not even be possible), you can use this to debug services in the cluster where these external services are already running.

plasne:
To be honest, I don't have enough interest in this to pursue it any further. We created this PR as part of a hack and it was requested that we post the findings here. Clearly that isn't the purpose of this playbook and so I will just cancel this PR.

PR #455, opened 2021-01-05
Added self-hosting agent recommendation to ui testing doc

abubinski:
Recommended self-hosted agent for running selenium tests in ADO.


RichCzyzewski (approved):

engineersamuel (approved):

engineersamuel:
Good update here.  I guess it all depends on what the build in packages are for the ms hosted agents which can be seen at https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml.   If installing anything requires sudo, for example on linux, then sudo isn't available (we found that out on a recent engagement).  That can be hacked around, but sometimes a self-hosted agent is easier!

alkofahi (approved):

PR #462, opened 2021-01-07
Missing eng-fundamentals-checklist link

cteicher-m:


omusavi (approved):

TessFerrandez (changes requested):
A couple of things here
1. This adds the content back which is great as it was removed earlier, but the link from the main doc is still missing
2. The issue being resolved by this has a few more tasks in it. I will change this issue to only be about adding back the engineering fundamentals checklist and adding the link to it from main.

TessFerrandez (approved):

PR #519, opened 2021-02-11
Remove Terraform links from CI IaC bullets

jskulavik:


TessFerrandez:
Why are we removing these links? is there an open issue or can you provide some context?

jskulavik:
We need to remove all references to and notions of the third-party Terraform tool given that it does not align with the goals of this playbook, our code-with practice, or our engineering fundamentals.

erikschlegel (commented):
The intent for the original section for IAC is that developers have options beyond ARM. If a customer is already committed to using terraform then we should meeting them where they are.

TessFerrandez:
So then it sounds @erikschlegel that we both think that we should leave the content as is.
@jskulavik if so, my suggestion is that we close this PR without merging, but would like to hear back from you

TessFerrandez (changes requested):
Suggesting to cancel this PR based on the comments above

fnocera:
@jskulavik suggest that we have a meeting (as we discussed over IM) and make sure we are aligned with @brockneedscoffee as well who is coordinating the CI/CD content. Will close this PR for now.

PR #534, opened 2021-02-24
DevEx - Working with local CI

itye-msft:
Added a new page which suggests a method to make your CI work locally as well in order to enrich the dev experience.


developer-experience/recipes/execute-local-pipeline.md
```diff
@@ -0,0 +1,32 @@
...
+
+1. Dockerize your application(s), including a build step if possible.
+2. Add a step in your docker file to execute unit tests.
+3. Add a step in the docker file for linting.
+4. Create a new dockerfile, possibly in a different folder, which executes end-to-end tests against the cluster. Make sure the default endpoints are configurable (This will become handy in your remote CI server, where you will be able to test against a live environment, if you choose to).
+5. Create a docker-compose file which allows to choose which of the services to run. The default will run only the applications, and an additional parameter can enable e2e tests for example.
+
+### Examples
+
+TODO: Add dockerfile and docker-compose samples, plus Azure Pipeline yaml to use it as a CI and a set of script commands to use it locally, to show how you can use the same resources locally and in the CI.
```
brockneedscoffee, line 31: should we put that content in the CI area and link it here?

developer-experience/recipes/execute-local-pipeline.md
```diff
@@ -0,0 +1,32 @@
+# Execute Local Pipeline
```
charleszipp, line 1: Lets rename this recipe to indicate its specific to docker and containers. How about `execute-local-pipeline-with-docker.md`?

developer-experience/recipes/execute-local-pipeline.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Method of work
+
+1. Dockerize your application(s), including a build step if possible.
+2. Add a step in your docker file to execute unit tests.
+3. Add a step in the docker file for linting.
+4. Create a new dockerfile, possibly in a different folder, which executes end-to-end tests against the cluster. Make sure the default endpoints are configurable (This will become handy in your remote CI server, where you will be able to test against a live environment, if you choose to).
+5. Create a docker-compose file which allows to choose which of the services to run. The default will run only the applications, and an additional parameter can enable e2e tests for example.
+
+### Examples
```
charleszipp, line 29: Are there any existing public repositories that may have a complete example of this? If not a potential follow up may be to make one that demonstrates the pattern.

developer-experience/recipes/execute-local-pipeline.md
```diff
@@ -0,0 +1,32 @@
...
+Having the ability to execute pipeline activities locally has been identified as an opportunity to promote positive developer experience.
+In this document we will explore a solution which will allow us to have the local CI experience to be as similar as possible to the remote process in the CI server.
+
+Using the suggested method will allow us to:
+
+1. Build
+2. Lint
+3. Unit test
+4. E2E test
+5. Deploy
```
charleszipp, line 14: nit: i would suggest to reword this to maybe Run Solution or something like that. In the rest of the content, i got the impression this was more geared towards how we could get services running so we run tests, etc. This doesnt seem to require us to "deploy" a container to AKS, ACI, or similar (which is actually a good thing!).

charleszipp (commented):
This is a good start. I believe the examples will really help drive the points and steps you made in the doc. This is a pattern ive seen a few crews use (including ours) so believe it will be helpful to have this added.

Another follow up content suggestion may be to work with @brockneedscoffee on something that pairs with this but is more from the ADO yaml perspective. I think you called this out in your TODO if i understand it right. How do we get our ADO pipeline to perform ci the same way a dev would locally. How do we use essentially the same source (i.e. docker files and bash) to run both to avoid duplicate code. Dont think its necessary to explain in this doc but would make for good follow up content in another area

charleszipp:
@itye-msft also, dont forget to address the markdown linter issues. there should be instructions for how to run the linter locally in contributing guidelines.

charleszipp (approved):

TessFerrandez (approved):

Issue #552, opened 2021-03-08
Add links to documented benefits of consolidated repos


TessFerrandez:
This comes from a TODO in the developer experience doc
EDIT: we should probably have pros and cons of both consolidated or separate repos

cteicher-m:
@TessFerrandez the source control section does not appear to have documented benefits of consolidated repos. Should this be added as well?

TessFerrandez:
This came from the developer experience section, yes, so the plan would be to add them in the source code section

@cteicher-m can you add some links or info? we've been talking about a section on mono vs. separate - I think it's not necessarily a one size fits all so we should probably then talk about pros and cons of both and where they fit.

TessFerrandez:
Closing due to lack of activity

Issue #551, opened 2021-03-08
Suggest hotkeys for mac in the Developer Experience Readme


TessFerrandez:
under established hotkeys

TessFerrandez:
Closed by PR

Issue #537, opened 2021-02-25
Engineering Checklist Definition: Security definition "Separation of Concerns"


dariuszparys:
As this came up in our team discussion, what exactly is meant with **Separation of Concerns**. Is the architectural pattern meant by this, if so why is it solely listed in the security section?

TessFerrandez:
@anthturner can you comment here

TessFerrandez:
Removing this as part of #553 but this issue can be repurposed to adding this back when we sort out what it means

PR #553, opened 2021-03-08
Removing TODOs from the playbook

TessFerrandez:
As the folks at github put it
image

These TODOs have been there for months now without completing which makes the playbook look un-cared for. Chances that they will ever be completed just by sitting there as TODOs are slimming. Therefore I have removed them and instead added any outstanding issues as issues in the playbook. 

- Rephrased the security items in the eng fundamentals checklist - and removed the "separation of concerns" checklist as I didn't know what it meant in the context of security - (checking in with @anthturner on this - but there is also an issue #537 open to address this and move it back when we get the answer)
- Removed content todos in the backlog refinement readme - created issue #554 
- Removed the todos in the developer experience section and created the issues #551 and #552 instead


deniscep (approved):

TessFerrandez:
Added wording for the separation of concerns (together with @anthturner)

tompaana (approved):

Issue #477, opened 2021-01-08
Add markdown link checking to the markdown recipe


TessFerrandez:
https://github.com/marketplace/actions/markdown-link-check

to https://github.com/microsoft/code-with-engineering-playbook/blob/master/code-reviews/recipes/Markdown.md

PR #589, opened 2021-03-18
add link check  to markdown recipe

katyamust:
[#477]
Add markdown link checking to the markdown recipe

### DIFF
--- a/.github/workflows/markdownlint.yml
+++ b/.github/workflows/markdownlint.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches:
       - master
-      
+
 jobs:
   lint:
     runs-on: ubuntu-latest
@@ -26,3 +26,4 @@ jobs:
       with:
         use-quiet-mode: 'yes'
         use-verbose-mode: 'no'
+        config-file: '.markdownlinkcheck.json'
--- a/.github/workflows/markdownlint.yml
+++ b/.github/workflows/markdownlint.yml
@@ -4,12 +4,10 @@ on:
   pull_request:
     branches:
       - master
-
+      
 jobs:
   lint:
-
     runs-on: ubuntu-latest
-
     steps:
     - uses: actions/checkout@v2
     - name: Use Node.js
@@ -28,4 +26,3 @@ jobs:
       with:
         use-quiet-mode: 'yes'
         use-verbose-mode: 'no'
-        config-file: '.markdownlinkcheck.json'
--- a/code-reviews/recipes/Markdown.md
+++ b/code-reviews/recipes/Markdown.md
@@ -87,7 +87,7 @@ Write Good is also available as an [extension for VS Code](https://marketplace.v
 
 ## VS Code Extensions
 
-#### Write Good Linter
+### Write Good Linter
 
 The [`Write Good Linter Extension`](https://marketplace.visualstudio.com/items?itemName=travisthetechie.write-good-linter) integrates with VS Code to give grammar and language advice while editing the document.
 
--- a/code-reviews/recipes/Markdown.md
+++ b/code-reviews/recipes/Markdown.md
@@ -14,7 +14,7 @@ Using Markdown is different than using a WYSIWYG editor. In an application like 
 
 You can find more information and full documentation [here](https://www.markdownguide.org/).
 
-## Code Analysis / Linting
+## Linting
 
 Markdown has specific way of being formatted. It is important to respect this formatting, otherwise some interpreters which are strict won't properly display the document. Linters are often used to help developers properly create documents by both verifying proper markdown syntax as well as grammar and proper English language.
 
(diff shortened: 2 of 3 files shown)


code-reviews/recipes/Markdown.md
```diff
@@ -85,17 +85,19 @@ npx write-good *.md
 
 Write Good is also available as an [extension for VS Code](https://marketplace.visualstudio.com/items?itemName=travisthetechie.write-good-linter)
 
-## VS Code Extensions
+### VS Code Extensions
```
TessFerrandez, line 88: This should stay a level 2 title to follow the same format as the other recipes

TessFerrandez, line 88:
```suggestion
## VS Code Extensions
```

code-reviews/recipes/Markdown.md
```diff
@@ -85,17 +85,19 @@ npx write-good *.md
...
 
-### Write Good Linter
+#### Write Good Linter
```
TessFerrandez, line 90:
```suggestion
### Write Good Linter
```

code-reviews/recipes/Markdown.md
```diff
@@ -85,17 +85,19 @@ npx write-good *.md
...
 
 The [`Write Good Linter Extension`](https://marketplace.visualstudio.com/items?itemName=travisthetechie.write-good-linter) integrates with VS Code to give grammar and language advice while editing the document.
 
-### markdownlint extension
+#### markdownlint extension
```
TessFerrandez, line 94:
```suggestion
### markdownlint extension
```

.github/workflows/markdownlint.yml
```diff
@@ -28,4 +26,3 @@ jobs:
       with:
         use-quiet-mode: 'yes'
         use-verbose-mode: 'no'
```
TessFerrandez, line 28: you need to keep the config-file: this contains the exceptions like ./link-to-the-work-item - this is why you are getting the link error in your PR.

TessFerrandez (changes requested):

code-reviews/recipes/Markdown.md
katyamust: I have got impression by reading the document that we are talking about linting tools here and VS code extensions are one of those tools - hence the change to level of title. Did I mistunderstood? @TessFerrandez

.github/workflows/markdownlint.yml
```diff
@@ -4,12 +4,10 @@ on:
   pull_request:
     branches:
       - master
-
+
```
tompaana, line 7:
```suggestion

```

code-reviews/recipes/Markdown.md
```diff
@@ -85,17 +85,19 @@ npx write-good *.md
...
 
 The [`markdownlint extension`](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) examines the markdown documents, showing warnings for rule violations while editing.
 
-## Build validation
+## Build Validation
+
+### linting
```
tompaana, line 100:
```suggestion
### Linting
```

code-reviews/recipes/Markdown.md
```diff
@@ -127,6 +129,20 @@ jobs:
         markdownlint "**/*.md" --ignore node_modules
 ```
 
+### link check
```
tompaana, line 132:
```suggestion
### Checking Links
```

tompaana (changes requested):

.github/workflows/markdownlint.yml
katyamust: Yepp, reverted back

tompaana (approved):
I see my suggested changes have been implemented. Approving on my behalf. Thank you!

code-reviews/recipes/Markdown.md
TessFerrandez: I think we should be consistent between recipes, the VS code extensions could be linting tools, they could also be other extensions that are useful - e.g. viewers, or in this case the write good extension etc. that are not necessarily linters

katyamust: Yes, that sound good then,  updated.

TessFerrandez (approved):

TessFerrandez:
@katyamust can you add the link that is failing to the exclude list? It should technically not fail the link check but seems to fail for all PRs for some reason so better to add it to the exclude list

katyamust:
> @katyamust can you add the link that is failing to the exclude list? It should technically not fail the link check but seems to fail for all PRs for some reason so better to add it to the exclude list

I have done that @TessFerrandez , thanks for tip. That link was actually working for me so I was not sure what to do. Now there are number of other dead links. This is becoming an issue I am fixing those almost every time  I submit something. 

May be we should tackle this differently and run link check only on timely basis. Lets discuss.

PR #622, opened 2021-04-28
Enable mega-linter in build

omusavi:
# Pull Request Template

Mega-Linter is a way to combine all of our current linters into a single toolchain and potentially build add more to ensure the quality of the playbook. Currently the only linters enabled are markdownlint and markdown-link-checker to match the current behavior in `main`.

One of the advantages of the mega-linter is it easily supports fixing issues that linters have caught, and with the current build rules auto-commits it to the current branch.

Future work could include enabling spell-check.

## Description of new changes

- Removed package.json since it is no longer required to run markdownlint locally (can be done with mega-linter-runner or VS Code)
- Added mega-linter workflow and configuration file

## For all pull requests

- [X] Label the pull request with the appropriate area(s)
- [X] Assign potential reviewers (you may also want to contact them on Teams to ensure timely reviews)
- [X] Assign the project - Engineering Playbook Backlog
- [X] Assign the pull request to yourself


.mega-linter.yml
```diff
@@ -0,0 +1,15 @@
...
+APPLY_FIXES: all # all, none, or list of linter keys
+DEFAULT_BRANCH: main
+SHOW_ELAPSED_TIME: true
+ENABLE:
+  - MARKDOWN
+  # - SPELL Currently 555 errors but would be nice to enable!
+DISABLE_LINTERS:
+  - MARKDOWN_MARKDOWN_TABLE_FORMATTER   # Does not correctly support GFM
+
+# Linter specific settings
```
omusavi, line 14: A couple ideas for settings we can put here to alleviate the markdown-link-checker issues:

- `MARKDOWN_MARKDOWN_LINK_CHECK_DISABLE_ERRORS: true` - show markdown link errors as warnings
- `MARKDOWN_MARKDOWN_LINK_CHECK_DISABLE_ERRORS_IF_LESS_THAN: 2` - we can set it to only error if we get more than 2 errors
- Configure `.markdownlinkcheck.json` to ignore all external links and only validate internal links

omusavi: Can't even get a clean CI build due to the link checker :(

fnocera: I think setting them to less than 2 errors seems very sensible - the url that your PR is blocked on works when navigating directly to it in the browser https://www.softwaretestinghelp.com/what-is-integration-testing/ so it's being blocked from being tested programmatically.

fnocera:
The linter overall takes c4min vs now we are at 2min. Just wanted to raised the increased time taken for the increased functionality/checking.

omusavi:
@fnocera most of that time is with the markdown link check, which looks like it can be pretty variable. [Here is another](https://github.com/microsoft/code-with-engineering-playbook/actions/runs/774277766) run that took roughly 4 minutes as well. But that is definitely something to monitor! FWIW looks like image pull time times are roughly the same

.mega-linter.yml
omusavi: Yeah I think that might be a good way to reduce the intermittency there, though it could potentially still let things leak through if the links we expect to fail actually pass and then an internal link is broken! Tradeoffs :) Happy to go either way

TessFerrandez (changes requested):
In general, I don't have anything against switching to this linter instead of the link checker we had before if it gives us benefits. I do however think (as mentioned in the comments above and in teams) that we should solve the root cause (i.e. disable checking of specific problematic links that give us false positives) rather than for example setting the limit to <2 links. This is not a sustainable solution, and this could also let real problems through.

I am also a little concerned about the added time that @fnocera mentioned.

.mega-linter.yml
TessFerrandez: Please merge in the current main branch to fix the link that fails

omusavi:
@TessFerrandez agreed, lets not set a acceptable failure count. Will merge from main and see if we can get a clean build and re-evaluate build times on that

.mega-linter.yml
omusavi: done, and passing!

brockneedscoffee:
@omusavi @fnocera @fnocera have we considered just having these things as pre-commit hooks to prevent things that would not pass our GH actions from being committed in the first place?  I don't see a reason why we could not just add that to the repo

omusavi:
@brockneedscoffee this is just my personal opinion but I do not like git commit/push hooks; they slow down the dev inner-loop too much and I prefer to let the CI system handle that. Especially if we do want to add more linters (like spellcheck), then that time would just increase. It also requires that everyone have these tools installed on their dev machine, which I like to always keep as clean as possible

zmmille2:
This is great. Please mention how to run it locally (is there a link to some docs we don't have to own?) in the `CONTRIBUTING.md`, I'm not sure how to run this otherwise.

zmmille2 (changes requested):
[This PR](https://github.com/microsoft/code-with-engineering-playbook/pull/682) will have to be essentially reverted if we do decide to move forward with mega linter.

TessFerrandez:
This PR has been stagnant for a while - do we want to revise and go forward @omusavi @fnocera @zmmille2, or just abandon?

TessFerrandez:
As there was no response to my question regarding whether we should abandon this PR - I will close it for now, but feel free to re-open at a later time if we want to incorporate it @omusavi @fnocera @zmmille2

PR #671, opened 2021-06-15
Describe an approach to use integration tests instead of unit tests

### LINKED ISSUE #669: Add an entry describing an approach to use integration tests instead of unit tests
**Is your feature request related to a problem? Please describe.**
In some situations integration tests can be used instead of unit tests to provide full test coverage.  Usually integration tests are used to verify that modules of a software system work together as designed and unit tests cover the functionality of isolated units of the system. In this approach the test coverage of integration tests is extended to verify the correct behaviour of the system as well.

**Describe the solution you'd like**
Add a paragraph about an approach for using integration tests instead of unit tests into “Unit testing” document in the playbook.

**Describe alternatives you've considered**
Adding the new entry to “Integration testing” section. But the context of “Unit testing” section seems to fit better because, in the described approach, integration tests act more like unit tests in a sense that they need to verify behaviour of the system. 

**Additional context**
Our team followed this approach successfully in the engagement.

Izzzu:
# Add an entry describing an approach to use integration tests instead of unit tests

In some situations integration tests can be used instead of unit tests to provide full test coverage. Usually integration tests are used to verify that modules of a software system work together as designed and unit tests cover the functionality of isolated units of the system. In the approach described in this PR the test coverage of integration tests is extended to verify the correct behaviour of the system as well.

Resolves #669 

- Add a paragraph about an approach for using integration tests instead of unit tests into “Unit testing” document in the playbook.
- Changes were made in "Unit testing" document: docs/automated-testing/unit-testing/README.md

## For all pull requests

- [x] Assign the project - Engineering Playbook Backlog


TessFerrandez (approved):

docs/automated-testing/unit-testing/README.md
```diff
@@ -179,6 +179,16 @@ Some common pitfalls when writing a unit test that are important to avoid:
   the system is not entirely testable. It is best to wrap third party API calls in interfaces or other structures so
   that they do not get invoked in unit tests.
 
+### Integration tests instead of unit tests
+
+In some situations it is worth considering to write integration tests instead of unit tests to provide full test coverage. The prerequisite for this approach to be successful is to have integration being able to execute at a speed comparable to that of unit tests both locally and in a CI environment.
```
marcgs, line 184:
```suggestion
In some situations it is worth considering to write integration tests instead of unit tests to provide full test coverage. The prerequisite for this approach to be successful is to have integration tests being able to execute at a speed comparable to that of unit tests both locally and in a CI environment.
```

docs/automated-testing/unit-testing/README.md
```diff
@@ -179,6 +179,16 @@ Some common pitfalls when writing a unit test that are important to avoid:
...
+
+Usually, integration tests prove that independently developed modules connect together as designed.
+In order to replace unit tests with integration tests, the test coverage of integration tests need to be extended to verify the correct behavior of the system as well. Instead of several unit tests needed to test a specific case of functionality of the system, one integration scenario is created that covers the entire flow. For example in case of an API, the received HTTP responses and their content are verified for each request in test. This covers both the integration between components of the API and the correctness of its business logic.
+
+Using efficient integration tests as a replacement for unit tests has the advantage of testing the system as a black box without any knowledge of its internals. Code refactoring has no impact on tests. Common testing techniques as TDD can be applied at a higher level which results in a development process that is driven by acceptance tests. Complex business logic and corner cases can be still tested with unit tests if necessary to prevent over-bloating the integration test suite.
```
alkofahi, line 190: I like the black box mentality, changing the environment to test and see how the service would behave.

docs/automated-testing/unit-testing/README.md
```diff
@@ -179,6 +179,16 @@ Some common pitfalls when writing a unit test that are important to avoid:
...
+
+In some situations it is worth considering to write integration tests instead of unit tests to provide full test coverage. The prerequisite for this approach to be successful is to have integration being able to execute at a speed comparable to that of unit tests both locally and in a CI environment.
+Most modern web frameworks like ASP.NET or Spring Boot combined with the right mocking or stubbing approach for external dependencies offer excellent capabilities to enable such scenarios for testing.
+
+Usually, integration tests prove that independently developed modules connect together as designed.
+In order to replace unit tests with integration tests, the test coverage of integration tests need to be extended to verify the correct behavior of the system as well. Instead of several unit tests needed to test a specific case of functionality of the system, one integration scenario is created that covers the entire flow. For example in case of an API, the received HTTP responses and their content are verified for each request in test. This covers both the integration between components of the API and the correctness of its business logic.
```
alkofahi, line 188: +1 to verifying the correct behavior. In your project, how did you scope/decide which behavior is to be verified? Did you go for happy and failure paths?

alkofahi (commented):
Hi folks, this is a thought-provoking contribution and I like it! Here is my concern though, if you think of code execution overall, each entry point (i.e. API) is a tree of execution paths, in those trees, each branch corresponds to a certain set of execution/code conditions to that entry point. Verifying the behavior for each of those paths might be infeasible as the number of paths grows exponentially with the number of conditions. 

This is one of the major challenges in relying on integration testing without unit testing. The nice thing about unit testing is that it allows us to test on a more granular and well-scoped level of the codebase, typically a method within a class, which might be feasible to test and verify the behavior of the branches/paths in that method.

I am not saying that we cannot replace UT with integration, I believe that can happen under certain conditions, for example, for simple services or with the owners acknowledging the risks of having one vs both, among others. My main feedback here is that we are pitching a replacement but without diving deep into the thoughts behind it and when it would be applicable. We need to avoid misguiding our readers into skipping UT in favor of adding integration tests without fully understanding when that is applicable.

I will be gone starting next week, plz close the loop on this review with Omeed

docs/automated-testing/unit-testing/README.md
```diff
@@ -179,6 +179,16 @@ Some common pitfalls when writing a unit test that are important to avoid:
   the system is not entirely testable. It is best to wrap third party API calls in interfaces or other structures so
   that they do not get invoked in unit tests.
 
+### Integration tests instead of unit tests
+
+In some situations it is worth considering to write integration tests instead of unit tests to provide full test coverage. The prerequisite for this approach to be successful is to have integration being able to execute at a speed comparable to that of unit tests both locally and in a CI environment.
+Most modern web frameworks like ASP.NET or Spring Boot combined with the right mocking or stubbing approach for external dependencies offer excellent capabilities to enable such scenarios for testing.
```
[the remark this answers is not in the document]
marcgs: I think we are saying "it depends". External dependencies that allow for efficient integration testing might not need to be mocked. An example of this could be an SQL database, instead of mocking the data layer of an application, one could use a local database with some test data in it, leading to a simpler test fixture setup than with mocking. Other more complex dependencies might be best mocked out completely or partially. Generally the mocking strategy needs to be evaluated carefully on a case by case basis to achieve a good trade-off of testing bigger software units efficiently.

marcgs:
> Hi folks, this is a thought-provoking contribution and I like it! Here is my concern though, if you think of code execution overall, each entry point (i.e. API) is a tree of execution paths, in those trees, each branch corresponds to a certain set of execution/code conditions to that entry point. Verifying the behavior for each of those paths might be infeasible as the number of paths grows exponentially with the number of conditions.
> 
> This is one of the major challenges in relying on integration testing without unit testing. The nice thing about unit testing is that it allows us to test on a more granular and well-scoped level of the codebase, typically a method within a class, which might be feasible to test and verify the behavior of the branches/paths in that method.
> 
> I am not saying that we cannot replace UT with integration, I believe that can happen under certain conditions, for example, for simple services or with the owners acknowledging the risks of having one vs both, among others. My main feedback here is that we are pitching a replacement but without diving deep into the thoughts behind it and when it would be applicable. We need to avoid misguiding our readers into skipping UT in favor of adding integration tests without fully understanding when that is applicable.
> 
> I will be gone starting next week, plz close the loop on this review with Omeed



> Hi folks, this is a thought-provoking contribution and I like it! Here is my concern though, if you think of code execution overall, each entry point (i.e. API) is a tree of execution paths, in those trees, each branch corresponds to a certain set of execution/code conditions to that entry point. Verifying the behavior for each of those paths might be infeasible as the number of paths grows exponentially with the number of conditions.
> 
> This is one of the major challenges in relying on integration testing without unit testing. The nice thing about unit testing is that it allows us to test on a more granular and well-scoped level of the codebase, typically a method within a class, which might be feasible to test and verify the behavior of the branches/paths in that method.
> 
> I am not saying that we cannot replace UT with integration, I believe that can happen under certain conditions, for example, for simple services or with the owners acknowledging the risks of having one vs both, among others. My main feedback here is that we are pitching a replacement but without diving deep into the thoughts behind it and when it would be applicable. We need to avoid misguiding our readers into skipping UT in favor of adding integration tests without fully understanding when that is applicable.
> 
> I will be gone starting next week, plz close the loop on this review with Omeed

You are right that the more classes we group together to test as a integration unit, the more execution paths that need to be covered. Testing complexity increases exponentially with the amount of elements we group together for integration testing.

However classes and methods are often not best suited as testing units. Often these are low level software elements, not representing any concept that can be grasped from a user perspective. Think of a controller class in a web application, normally the job of such classes is just to orchestrate calls to lower level collaborators. Writing a unit test for such class normally leads to mocking every collaborator and verifying that the right calls are being done. We end up duplicating our code in the test to achieve that.

The idea we are trying to promote is that often is best to treat higher level components with several classes interacting together as our testing units. This allows to verify behaviour at a higher level, without verifying any specifics of the implementation underneath. In case of a web api these components could be its HTTP endpoints. You are right that sometimes the testing complexity will be too high, and that is when a combination of both techniques will come handy, quoting our last paragraph "Complex business logic and corner cases can be still tested with unit tests if necessary to prevent over-bloating the integration test suite".

Looking forward to further thoughts and experiences in this area :-)

docs/automated-testing/unit-testing/README.md
Izzzu: Yes, we went for both success and failure scenarios. Our approach was basically to cover all use cases using integration tests.

docs/automated-testing/unit-testing/README.md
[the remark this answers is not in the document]
Izzzu: In our case it was indeed both branch and line coverage, but the needed test coverage can depend on the projects specifics. So the “full test coverage” in this case can be misleading. What we meant is “sufficient test coverage to ensure that the system is working properly”.

omusavi:
> However classes and methods are often not best suited as testing units. Often these are low level software elements, not representing any concept that **can be grasped from a user perspective.** 

This right here is where I think the main difference in opinion comes from. Unit tests are explicitly not designed to be from the users perspective. They are at their core, developer tests. Like you mentioned, as you add more classes, integration testing becomes exponentially more difficult which also makes the overall test suite much harder to maintain. 

Ultimately, I really like your approach for integration testing. Integration tests that run quickly and can even be run in the dev environment fairly easily would make my life as a developer much easier, and encourage me to write integration tests. They also differentiate themselves from E2E tests well in that E2E would be more "live" whereas integration is trying to isolate sets of related components and testing their interactions. But I think that this is an approach that we should encourage for ALL integration tests and not just a replacement for unit tests.

> Writing a unit test for such class normally leads to mocking every collaborator and verifying that the right calls are being done. We end up duplicating our code in the test to achieve that.

I have 3 issues with this. If you have so many dependencies in your controllers that you have a tough time mocking all of them out that is a code smell. One of the things that unit testing does that I do not believe your strategy with integration testing does as well is that it by writing unit tests, it helps dictate a testable design. When you see a single class with 5-6 different dependencies, that should be a clue to you that this one class is likely doing too much and is not fitting the single responsibility principle.

Secondly, and this is more of an opinion, I am not a believer in mocking and instead am a strong proponent in using simulators or fakes for your dependencies. These fakes can be checked in and co-located with product code instead of isolated to test code, for example an in-memory implementation of a DB. Many of your problems with unit testing are actually resolved by not using mocks and instead using these simulators which mimic real world actions. And in many cases I find I end up using these simulators in the product code sometimes! Simulators also can be unit tested themselves and those tests can act as a way of testing an interface rather than testing an actual implementation. 

Thirdly, you mention that you will be verifying that the right calls are being done. This is typically called Behavior validation and is something very common with mocks. I generally am against behavior validation because it is so tied to a specific implementation. For example, if I write a test that validates that I make a certain call 3 times, then later I make it more efficient and only call it once, now my test fails. But that indicates to me that the test was written poorly in the first place. The unit test CAN be black-box, and in a lot of ways should be as well. You are testing the contract of the unit and that the final state of the unit is changed in the way that you expect it to me, and not its behavior.

---

My main concern is that someone will be going through the code-with playbook, see this section and decide "oh i'll just do integration testing" because let's be honest, writing good unit tests is hard. But it is also virtuous and a good unit testing suite will save time and energy down the road. What I would recommend is that we take the current section you all have written and work in these learnings into the page on integration tests. I think you all have some really good points and strategies on how to do integration tests well and I would love to see that added to our playbook. Maybe even we include a section in the unit test page explaining where unit testing is not as useful and suggest integration testing for those occasions (like, for example the SQL DB interactions you mentioned which would likely be completely uncovered in UTs)

Just my thoughts on that, let me know what you think. Also would be happy to jump on a call if that is easier!

Izzzu:
> I have 3 issues with this. If you have so many dependencies in your controllers that you have a tough time mocking all of them out that is a code smell. One of the things that unit testing does that I do not believe your strategy with integration testing does as well is that it by writing unit tests, it helps dictate a testable design. When you see a single class with 5-6 different dependencies, that should be a clue to you that this one class is likely doing too much and is not fitting the single responsibility principle.

> Secondly, and this is more of an opinion, I am not a believer in mocking and instead am a strong proponent in using simulators or fakes for your dependencies. These fakes can be checked in and co-located with product code instead of isolated to test code, for example an in-memory implementation of a DB. Many of your problems with unit testing are actually resolved by not using mocks and instead using these simulators which mimic real world actions. And in many cases I find I end up using these simulators in the product code sometimes! Simulators also can be unit tested themselves and those tests can act as a way of testing an interface rather than testing an actual implementation.

> Thirdly, you mention that you will be verifying that the right calls are being done. This is typically called Behavior validation and is something very common with mocks. I generally am against behavior validation because it is so tied to a specific implementation. For example, if I write a test that validates that I make a certain call 3 times, then later I make it more efficient and only call it once, now my test fails. But that indicates to me that the test was written poorly in the first place. The unit test CAN be black-box, and in a lot of ways should be as well. You are testing the contract of the unit and that the final state of the unit is changed in the way that you expect it to me, and not its behavior.

Thank you for your feedback and very valuable points! 

In general our goal here was to show that integration tests can have more responsibility than only increasing a confidence in the development cycle, because they prove that independently developed modules interface as technically designed. As you mentioned writing good unit tests is hard, so it’s easy to do it incorrectly, which can lead to code smells or using some inefficient technics. With the integration tests approach the rules are, in our opinion, easier to follow, as the main focus here is on the happy/failure paths in the system, that can be tested from the outside. It’s possible with the frameworks that we also use for unit testing and that’s why it makes the integration tests so easy from the development perspective. 
Testing with unit tests as a black box seems to be a bit different as we still have to know the internals of the system. With integration tests the black box is on the higher level, which allows to have a freedom with choosing the right approach to the design of the internals. In this approach unit tests can still be used as a tool to dictate a testable code, but they don’t need to also have the responsibility of providing the sufficient coverage, as all the use cases are covered already by integration tests, which become acceptance tests. 

I modified the last paragraph to stress also, that unit tests can play an important role in the system.

It would be great to have a call and discuss your suggestions and how we can incorporate the points into the playbook. Thank you!

Izzzu:
> > However classes and methods are often not best suited as testing units. Often these are low level software elements, not representing any concept that **can be grasped from a user perspective.**
> 
> This right here is where I think the main difference in opinion comes from. Unit tests are explicitly not designed to be from the users perspective. They are at their core, developer tests. Like you mentioned, as you add more classes, integration testing becomes exponentially more difficult which also makes the overall test suite much harder to maintain.
> 
> Ultimately, I really like your approach for integration testing. Integration tests that run quickly and can even be run in the dev environment fairly easily would make my life as a developer much easier, and encourage me to write integration tests. They also differentiate themselves from E2E tests well in that E2E would be more "live" whereas integration is trying to isolate sets of related components and testing their interactions. But I think that this is an approach that we should encourage for ALL integration tests and not just a replacement for unit tests.
> 
> > Writing a unit test for such class normally leads to mocking every collaborator and verifying that the right calls are being done. We end up duplicating our code in the test to achieve that.
> 
> I have 3 issues with this. If you have so many dependencies in your controllers that you have a tough time mocking all of them out that is a code smell. One of the things that unit testing does that I do not believe your strategy with integration testing does as well is that it by writing unit tests, it helps dictate a testable design. When you see a single class with 5-6 different dependencies, that should be a clue to you that this one class is likely doing too much and is not fitting the single responsibility principle.
> 
> Secondly, and this is more of an opinion, I am not a believer in mocking and instead am a strong proponent in using simulators or fakes for your dependencies. These fakes can be checked in and co-located with product code instead of isolated to test code, for example an in-memory implementation of a DB. Many of your problems with unit testing are actually resolved by not using mocks and instead using these simulators which mimic real world actions. And in many cases I find I end up using these simulators in the product code sometimes! Simulators also can be unit tested themselves and those tests can act as a way of testing an interface rather than testing an actual implementation.
> 
> Thirdly, you mention that you will be verifying that the right calls are being done. This is typically called Behavior validation and is something very common with mocks. I generally am against behavior validation because it is so tied to a specific implementation. For example, if I write a test that validates that I make a certain call 3 times, then later I make it more efficient and only call it once, now my test fails. But that indicates to me that the test was written poorly in the first place. The unit test CAN be black-box, and in a lot of ways should be as well. You are testing the contract of the unit and that the final state of the unit is changed in the way that you expect it to me, and not its behavior.
> 
> My main concern is that someone will be going through the code-with playbook, see this section and decide "oh i'll just do integration testing" because let's be honest, writing good unit tests is hard. But it is also virtuous and a good unit testing suite will save time and energy down the road. What I would recommend is that we take the current section you all have written and work in these learnings into the page on integration tests. I think you all have some really good points and strategies on how to do integration tests well and I would love to see that added to our playbook. Maybe even we include a section in the unit test page explaining where unit testing is not as useful and suggest integration testing for those occasions (like, for example the SQL DB interactions you mentioned which would likely be completely uncovered in UTs)
> 
> Just my thoughts on that, let me know what you think. Also would be happy to jump on a call if that is easier!

Thank you for your feedback and very valuable points! 

In general our goal here was to show that integration tests can have more responsibility than only increasing a confidence in the development cycle, because they prove that independently developed modules interface as technically designed. As you mentioned writing good unit tests is hard, so it’s easy to do it incorrectly, which can lead to code smells or using some inefficient technics. With the integration tests approach the rules are, in our opinion, easier to follow, as the main focus here is on the happy/failure paths in the system, that can be tested from the outside. It’s possible with the frameworks that we also use for unit testing and that’s why it makes the integration tests so easy from the development perspective. 
Testing with unit tests as a black box seems to be a bit different as we still have to know the internals of the system. With integration tests the black box is on the higher level, which allows to have a freedom with choosing the right approach to the design of the internals. In this approach unit tests can still be used as a tool to dictate a testable code, but they don’t need to also have the responsibility of providing the sufficient coverage, as all the use cases are covered already by integration tests, which become acceptance tests. 

I modified the last paragraph to stress also, that unit tests can play an important role in the system.

It would be great to have a call and discuss your suggestions and how we can incorporate the points into the playbook. Thank you!

omusavi (approved):
Thanks so much for making these changes, this is exactly what I had in mind! Thanks again for your contributions here

Izzzu:
> Thanks so much for making these changes, this is exactly what I had in mind! Thanks again for your contributions here

Thank you for your feedback!

PR #712, opened 2021-09-22
move decision-logs examples back into the design tree

TessFerrandez:
# Pull Request Template

The design reviews folder has moved under design, but when merging an old PR, some design review content was accidentally left in the root. This moves it to the right location


fnocera (approved):
Thanks for fixing this!

alkofahi (approved):
Thank you for taking care of this!

PR #748, opened 2021-11-09
rename section titles to render better on the site

Ethan-Arrowood:
This PR fixes some small issues with the way the UI UX section titles were rendered. Not sure if there is a way to override it since I do like "UI/UX". I'm not familiar with whatever static site generator we are using here.

Additionally, re-titled one of the documents so it renders better as well.

This screenshot demonstrates the current issues:
Screen Shot 2021-11-09 at 4 21 50 PM

I don't know how to run the site locally otherwise I would've caught this before merging my previous PR


TessFerrandez (approved):
@Ethan-Arrowood you can run it locally using mkdocs serve (need to first install mkdocs though -- if you have python installed you can do this with `pip install -r requirements-docs.txt` from the root dir)

Ethan-Arrowood:
I'll work on getting a dev container built so that folks can run the site locally. thank you!

wbreza (approved):
LGTM

Ethan-Arrowood:
@TessFerrandez i just did some digging on overriding the section title. Its quite a rough dev experience. Sort of all or nothing, as in if we want to change the name of UI / UX we have to manually type out all of the other navigation paths. I tried seeing if there was an alternative via the theme but i don't think its easily possible. Additionally, I tried seeing how hard it would be to implement a nav override so we could customize singular titles, but again the complexity there goes beyond my knowledge of mkdocs. 

All in all, I think leaving the title as is is the easiest path forward. No need to make it more complex just so it could display "UI / UX" vs "User Interface Engineering" 😁

PR #806, opened 2022-03-08
remove broken link

TessFerrandez:
- Removes broken link from the ML model checklist (other links that cover the same topic already exist so don't think we need to replace)


biancafbmd (approved):
Thanks for fixing this Tess :)

omri374 (approved):

Issue #858, opened 2022-06-30
Add recipe to explain how to share common variables between Terraform modules


jcorioland:
When deploying infrastructure from code, it's common practice to split the code into different modules responsible for the deployment of a part or component of the infrastructure. In Terraform, this can be done by using modules.

In such kind of situation, it is useful to be able to share some common variables as well as centralize naming conventions of the different resources in a central fashion, to ensure it's easy to refactor when it has to change, despite the dependencies that exist between modules.

This recipe will propose a solution to solve that problem.

jcorioland:
Closed by #859

PR #859, opened 2022-06-30
Add recipe for sharing common variables and naming between TF modules

jcorioland:
This PR solves #858.

# Pull Request Template

When deploying infrastructure from code, it's common practice to split the code into different modules responsible for the deployment of a part or component of the infrastructure. In Terraform, this can be done by using modules.

In such kind of situation, it is useful to be able to share some common variables as well as centralize naming conventions of the different resources in a central fashion, to ensure it's easy to refactor when it has to change, despite the dependencies that exist between modules.

## Description of new changes

This PR adds a recipe with a solution to solve the problem described above.

## For all pull requests

- [ ] Assign the project - Engineering Playbook Backlog


docs/continuous-delivery/recipes/terraform/share-common-variables-naming-conventions.md
```diff
@@ -0,0 +1,304 @@
...
+Using the common Terraform module in any other module is super easy. For example, this is what you can do in the Azure Kubernetes module `main.tf` file:
+
+```hcl
+module "common" {
+  source           = "../common"
+  environment_name = var.environment_name
+  subscription     = var.subscription
+}
+
+data "azurerm_subnet" "aks_subnet" {
```
omusavi, line 154: In the past when there is a single resource of that type, I have seen the usage of the name `this` for the resource name. Do you think this is something that is a convention we should also add to this doc? I am not sure if there is guidance for this from the solution area

Example:
```hcl
data "azurerm_subnet" "this" {
```

jcorioland: @omusavi this is a practice that I've seen a very few times in the past, but I find it confusing TBH. In most projects I've seen, data sources and resources that are created inside a module are considered as instance of objects in the module instance itself and named using very explicit nouns like it would have been done in C# or Python, for example.

@hattan does your team have some best practices recommendations for such a case?

jcorioland: BTW, according to [this page](https://github.com/microsoft/code-with-engineering-playbook/blob/main/docs/code-reviews/recipes/terraform.md) of the playbook, we shoud follow this [naming convention](https://github.com/jonbrouse/terraform-style-guide/blob/master/README.md#resource-naming) at CSE. And it recommends to use the same name as the type, minus the provider. So in this case, I should have use `subnet`. And `subnet_aks` if multiple subnets were referenced in the module.

meulta: @omusavi can you share the advantage of this? If it has more advantage than the current naming convention we should consider it

meulta (approved):
Thanks for the contribution! 

Looks good to me

docs/continuous-delivery/recipes/terraform/share-common-variables-naming-conventions.md
omusavi: no major advantage, but this is the naming convention i have always seen. Thanks for the link to the playbook page with the CSE opinion on convention. I think `this` is nice because it's always consistent, and since you always have to include the resource type anyways, it seems repetitive to me. Thanks for all the links though!

omusavi (approved):

jcorioland:
@omusavi @meulta thanks for the reviews & approvals.
It looks like I am not allowed to merge the PR myself in this repo. Can one of you press the button for me, please? :-)
Thanks !

Issue #870, opened 2022-08-19
[Automated Testing] Add best practices for mocking during unit tests


julyvee:
**Is your feature request related to a problem? Please describe.**
While testing is essential, it's often thought of too late and there isn't enough time to implement it properly. In our last engagement, we realized some ways how to save time while writing tests and making them more effective.

**Describe the solution you'd like**
I want to add some best practices for mocking to the unit testing section of the playbook.

**Describe alternatives you've considered**

**Additional context**
The learnings are currently published on a personal blog, here: https://julyvee.github.io/2022/08/17/dos-and-donts-for-efficient-unit-testing/

PR #872, opened 2022-08-22
Add best practices for mocking

julyvee:
# Pull Request Template

Adding some best practices for mocking, see issue https://github.com/microsoft/code-with-engineering-playbook/issues/870

## Description of new changes

- Added two best practices for mocking to the unit testing mocking section
- Reinforced the importance of dependency injection

## For all pull requests

- [x] Assign the project - Engineering Playbook Backlog


docs/automated-testing/unit-testing/mocking.md
```diff
@@ -92,16 +92,128 @@ since it is held to the same standard with full unit tests. Especially if writin
...
 - No framework needed, is just like any other implementation.
 - Encourages testable design.
 - Code can be "promoted" to product code, so it is not wasted effort.
 
-### Downsides
+### Downsides to Fakes
 
 - Takes more time to implement.
 
+## Best Practices
```
TessFerrandez, line 105: The need for dependency injection to do mocking is only relevant for certain languages (ex. C#) but many languages (ex. Python) doesn't need this, nor is it a common practice to do dependency injection. Might be worth mentioning this.

docs/automated-testing/unit-testing/mocking.md
```diff
@@ -92,16 +92,128 @@ since it is held to the same standard with full unit tests. Especially if writin
...
+
+To keep your mocking efficient, consider these best practices to make your code testable, save time and make your
```
TessFerrandez, line 107: These best practices suggest mocking and DI are our preferred ways to do unit testing. 

IMHO we want to try to design away the need for mocking in general, and perhaps consider integration tests for the glue code that has outside dependencies, and avoid doing testing theatre where we add unit tests to get 100% on glue code that has little logic and that would be better covered in integration tests.

But then of course in C# and the likes, when we can't design away the need for unit tests of glue code for whatever reason, we want to make sure that we design for testing with DI etc.

Thoughts?

docs/automated-testing/unit-testing/mocking.md
```diff
@@ -92,16 +92,128 @@ since it is held to the same standard with full unit tests. Especially if writin
...
+test assertions more meaningful.
+
+### Dependency Injection
+
+If you don’t keep testability in mind from the beginning, once you start writing your tests, you might realize you have
+to do a time-intensive refactor to make the code unit testable. A common problem that leads to non-testable code is not
+using dependency injection. Use dependency injection so that a mock can easily be injected into your SUT during a unit test.
```
TessFerrandez, line 114:
```suggestion
using dependency injection. Use dependency injection so that a mock can easily be injected into your Subject Under Test (SUT) during a unit test.
```

docs/automated-testing/unit-testing/mocking.md
julyvee: That's a good point! The example I used in the Assertions best practice is very simple on purpose to convey the idea but avoid having overly complex code snippets. I agree that not every piece of glue code should be unit tested, but if a unit test with mocks is required, the principles still apply. I suggest adding a disclaimer to clarify better in which situations these principles should be used. What do you think? Or would you prefer changing the example to have more logic?

docs/automated-testing/unit-testing/mocking.md
julyvee: Thanks for the feedback! I added a disclaimer about different languages :)

TessFerrandez (approved):

docs/automated-testing/unit-testing/mocking.md
julyvee: Added. Let me know if this is okay, or you prefer further disclaimers or changes to the example :)

PR #892, opened 2022-09-29
Enable xops integration

day-jeff:
# Pull Request Template

- We are including all CWEP content in the SolutionOps playbook. (Full details still to be worked out--this is a first step.)
- The SolutionOps playbook uses the markdownextractdata plugin, which uses jinja2.
- In order to publish CWEP pages into the SolutionOps playbook, they must be parsed correctly by all the plugins that are installed in the SolutionOps mkdocs.yml config.
- Several CWEP pages have yaml code snippets that use double curly braces (`{{ ... }}`), which is the jinja2 substituion operator, for parameters.
- The yaml snippets break `mkdocs build` when jinja2 parses them during the build.

## Description of new changes

- Add the markdownextractdata extension to the CWEP mkdocs.yml configuration, so that any build errors caused by yaml snippets are discovered in the source repo rather than on import to the SolutionOps repo.
- Add {% raw %} and {% endraw %} tagas aroudn yaml code snippets to disable jinja parsing. Assures the updated pages can be published correctly by into the SolutionOps playbook.

## For all pull requests

- [X] Label the pull request with the appropriate area(s)
- [X] Assign potential reviewers (you may also want to contact them on Teams to ensure timely reviews)
- [X] Assign the project - Engineering Playbook Backlog
- [X] Assign the pull request to yourself


TessFerrandez:
Before we start modifying the EF playbook for something that may or may not happen - can we have a discussion first?
I would suggest we hold off on this change until that discussion has happened.

TessFerrandez (changes requested):
see overall comment

TessFerrandez:
Should we close this PR at this point - I think the idea at the moment is that the play books will continue to be separate entities but with strong collaboration.

PR #908, opened 2022-11-18
adding checkly to synthetics tools

hlenke:


TessFerrandez (approved):

TessFerrandez:
#911 will fix the link error if we merge that one first and then this on top

PR #928, opened 2023-02-03
Modifying Stub Documentation With Python Example

mohanajuhi166:
# Modifying Stub's Description With A Python Specific Example

- We are currently modifying the stub's description to provide a more specific definition on how stubs works. This includes providing scenarios where stubs are useful
- Providing a python specific example for stubs section

## Description of new changes

- We are currently modifying  the lines involving S**tubs typically ignore inputs they do not expect or will always return the same thing** and replace it with a more specific definition of stubs . In this we have included 

-  what is a Stub
- how they work
- what kind of logic can be present in stub
- when are they useful

The last point (when are they useful) is key , because this helps to **use stubs in situations when we have to test APIs we dont own.**

We have also provided a python code sample for the same.

## For all pull requests

- [ ] Assign the project - Engineering Playbook Backlog
- [ ] Assign the pull request to yourself

### DIFF
--- a/docs/automated-testing/unit-testing/mocking.md
+++ b/docs/automated-testing/unit-testing/mocking.md
@@ -12,9 +12,12 @@ choice and Fakes should be considered instead.
 
 ## Stubs
 
-Most of the time when "mocks" are being used, what is actually being used is a **stub**; a stub simply returns what the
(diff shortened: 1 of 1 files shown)


TessFerrandez (approved):
Thank you for the contribution

shiranr (approved):
LGTM! Thank you :)

PR #931, opened 2023-02-06
Refactor Dev Ex

TessFerrandez:
The Developer Experience section has some weird artifacts

1. There is a page listed as Problem Statement
2. There is duplicate content in the recipes and the root of dev ex
3. There is no real distinction between what goes in the root and in recipes

This PR fixes the following

1. Removes the duplicate content
2. Adds a better title so it doesn't show up as problem statement
3. Removes the recipes folder and moves everything to the root

This is only refactoring, now actual changes to content other than the title, folder structure and white space


shiranr (approved):
LGTM

jchomarat (approved):
All good to me this re-organization

PR #944, opened 2023-03-07
Git commit and versioning improvements

LeonardHd:
# Pull Request Template

- Move commit best practices to git guidance
- adding conventional commit message structure to commit best practices
- fixing issue on semantic versioning and linking GitVersion to actually determine version from git history


shiranr (approved):

docs/source-control/component-versioning.md
```diff
@@ -30,31 +30,35 @@ Updating the major / minor version indicates changes of the API / interfaces / c
...
-Semantic Versioning is a concept of calculating the version number automatically based on a certain source code repository.
-
-The `semver` tool looks at a GIT source control branch and comes up with a *repeatable* and *unique* version number based on
-
-* number of commits since last major or minor release
-* commit messages
-* tags
-* branch names
+Semantic Versioning is a versioning scheme specifying how to interpret the different version numbers. The most common format is `major.minor.patch`. The version number is incremented based on the following rules:
+* Major version when you make incompatible API changes,
```
TessFerrandez, line 34:
```suggestion

* Major version when you make incompatible API changes,
```

docs/source-control/git-guidance/README.md
```diff
@@ -232,6 +232,74 @@ Then you can use the reflog reference (`HEAD@{}`) to reset to a specific commit
...
+A good commit message should answer these questions:
+
+* Why is it necessary? It may fix a bug, add a feature, improve performance, or just be a change for the sake of correctness
+* How does it address the issue? For short, obvious changes, this can be omitted
+* What effects does this change have? In addition to the obvious ones, this may include benchmarks, side effects etc.
+* What limitations does the current code have?
+
+Consider this when writing your commit message:
+
+* Don't assume that the code is self-evident/self-documenting
```
TessFerrandez, line 254: Not sure I agree with this... if the code is not self-evident, we should work to make it so. 

What we write in the PR will only be there while the PR is open - the code is the long-lasting result, and as such, any questions we may have about the code, should be answered by the code itself.

Can we re-phrase this, or remove it?

docs/source-control/git-guidance/README.md
```diff
@@ -232,6 +232,74 @@ Then you can use the reflog reference (`HEAD@{}`) to reset to a specific commit
 git reset HEAD@{2}
 ```
 
+## Commit Best Practices
```
TessFerrandez, line 235: This section has bothered me for a while, and I am wondering if we can maybe re-phrase it a bit.

To me, a commit is a ```git commit -m "some message"``` and I personally think we should encourage people to commit extremely often and with fairly brief commit messages - ex. "implement GET /todos endpoint"

The guidance given here, is, in my opinion more geared towards PR descriptions, where one PR may contain 40 commits.

Do you agree? And if so, should we simply move this out all together, and add the relevant parts of this content to the guidance for code reviews.

docs/source-control/git-guidance/README.md
```diff
@@ -232,6 +232,74 @@ Then you can use the reflog reference (`HEAD@{}`) to reset to a specific commit
...
+
+You can specify the default git editor, which allows you to write your commit messages using your favorite editor. The following command makes Visual Studio Code your default git editor:
+
+```bash
+git config --global core.editor "code --wait"
+```
+
+### Commit Message Structure
+
+You are free to structure commit messages. However, it can be helpful to follow a convention within your team.
```
TessFerrandez, line 266: I think all this content comes from a post on "how to write good commits" - maybe we can simply remove all of this and link to the post itself and thus make the documentation here a lot cleaner.

docs/source-control/git-guidance/README.md
TessFerrandez: PS! I know this was part of a re-factor, but since this PR is a mix of re-factoring and functional code changes in one, it makes it fair game to comment :)

TessFerrandez (commented):
Happy to discuss on teams if you want to have a longer discussion...

docs/source-control/git-guidance/README.md
LeonardHd: I moved this part from the original README.md
Personally, I rather say source code should be rather self-documenting then not (so I see your point here).

In fact, a commit message should rather not explain the source code but describe the change. I suggest to delete this line.

docs/source-control/git-guidance/README.md
LeonardHd: I reduced the content a little more and only added the most essential "rules" that the git command expects.

Do we want to keep the sections Commit Best Practices and Commit Message? They are linked in other documents so might be good to keep them. However, we might want to clean them up a little further.

In particular, the following could be compressed into one:
```text
A good commit message should answer these questions:

* Why is it necessary? It may fix a bug, add a feature, improve performance, or just be a change for the sake of correctness
* How does it address the issue? For short, obvious changes, this can be omitted
* What effects does this change have? In addition to the obvious ones, this may include benchmarks, side effects etc.
* What limitations does the current code have?

Consider this when writing your commit message:

* If it seems difficult to summarize your commit, it may be because it includes more than one logical change or bug fix. If so, it is better to split it into separate commits with `git add -p`
* Don't assume the reviewer understands the original problem. It should be possible to review a change request without reading the contents of the original bug/task.

```

TessFerrandez: If we remove them we can just remove the links from the other sections
I'll approve the PR and leave it up to you how you want to finish it up

TessFerrandez (approved):

PR #959, opened 2023-04-11
Terraform structure guidelines

dianaivan11:
# Pull Request Template

This PR includes a set of best practices on how to structure a Terraform infrastructure configuration in order for it to be maintainable, scalable and reusable. 

The best practices presented here were extracted as part of the sharing phase from a recent CSE engagement, in which our crew leveraged the configuration structure previously used in another engagement. Since these guidelines were already being shared without being documented, including them in the EF playbook will enable more crews to leverage them in their projects.

### DIFF
--- a/docs/continuous-delivery/recipes/terraform/terraform-structure-guidelines.md
+++ b/docs/continuous-delivery/recipes/terraform/terraform-structure-guidelines.md
@@ -1,0 +1,69 @@
+# Guidelines on Structuring the Terraform Configuration
+
+## Context
+When creating an infrastructure configuration, it is important to follow a consistent and organized structure to ensure maintainability, scalability and reusability of the code. The goal of this section is to briefly describe how to structure your Terraform configuration in order to achieve this.
+
+## Structuring the Terraform configuration
+
+The recommended structure is as follows:
+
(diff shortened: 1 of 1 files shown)


TessFerrandez (approved):

docs/continuous-delivery/recipes/terraform/terraform-structure-guidelines.md
```diff
@@ -0,0 +1,69 @@
+# Guidelines on Structuring the Terraform Configuration
+
+## Context
+When creating an infrastructure configuration, it is important to follow a consistent and organized structure to ensure maintainability, scalability and reusability of the code. The goal of this section is to briefly describe how to structure your Terraform configuration in order to achieve this.
+
+## Structuring the Terraform configuration
+
+The recommended structure is as follows:
+
+1. Place each component you want to configure in it's own module folder. This will give you a clear separation of concerns and will make it straight forward to include new resources, update existing ones or reuse them in the future.
```
stuartleeks, line 10: How is a component defined in this context?

My concern is that it could be interpreted as an Azure resource, which can make the terraform hard to understand.

The terraform docs have some good high-level content on [when to write a module](https://developer.hashicorp.com/terraform/language/modules/develop#when-to-write-a-module) which might also be worth referencing here?

shiranr (approved):

PR #1045, opened 2024-04-12
Fix linter

shiranr:


nyouens (approved):

tompaana (approved):

shiranr:
Thank you @tompaana and @nyouens for the quick reply :)

PR #1065, opened 2024-08-22
Fix page navigation

TessFerrandez:
Very small changes to fix the navigation of the EF playbook


TessFerrandez:
Quick merge to fix navigation - OK per EF group

PR #1075, opened 2024-09-27
Tessferrandez/refactor ml to apply to all ai ml projects

TessFerrandez:
Based on feedback that the ML section seems to apply only to data scientists, this is an attempt to make the section more general as many projects include AI, and this content should apply to all of us.

Looking for feedback how we can make it even more accessible...


README.md
```diff
@@ -63,7 +63,7 @@ A [breakdown of sections](docs/the-first-week-of-an-ise-project.md) according to
 
 ## Fundamentals for Specific Technology Areas
 
-* [Machine Learning Fundamentals](docs/machine-learning/README.md)
+* [Machine Learning/AI Fundamentals](docs/machine-learning-and-ai-projects/README.md)
```
shiranr, line 66: Perhaps ml-and-ai-projects?

PR #1083, opened 2025-01-22
Clarify TDD doesn't mean write *all* tests first

joshua-koehler:
This was already written well, and the separate article `./tdd.md` is excellent. However, I think the language could be a little clearer to dispell the common misconception that TDD means you write all your tests first.  I trust this is clear to TDD practitioners, but I commonly see "senior" engineers misunderstand this.

Kent Beck sees this so commonly that he wrote a [little article](https://tidyfirst.substack.com/p/canon-tdd) on the subject.

My org was using this document for training engineers, and am hoping this small change will make it even more helpful.

I also made a minor change to the first sentence.  I understand it to be conveying one benefit of TDD, namely that it makes the code more testable, and thought it could be made more succinct.  Furthermore, many tout TDD as a mechanism for design, so explicitly negating that benefit is perhaps likely to cause unintended controversy.

### DIFF
--- a/docs/automated-testing/unit-testing/README.md
+++ b/docs/automated-testing/unit-testing/README.md
@@ -97,11 +97,12 @@ An example of using dependency injection can be found [here](./authoring-example
 
 #### Test-Driven Development
 
-Test-Driven Development (TDD) is less a technique in how your code is designed, but a technique for writing your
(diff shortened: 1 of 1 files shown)


nyouens (approved):
This is a great iteration on the content, thank you for the contribution.

shiranr (approved):

PR #1119, opened 2025-12-22
Added an E2E test tool 

falcon027:
Added an E2E tool that works well with web apps such as Microsoft Dynamics 365. Called kogiQA


falcon027:
@microsoft-github-policy-service agree

PR #1136, opened 2026-07-21
fix: enable CI checks to run from devcontainer

vdstrizhkova:

This PR makes all CI checks runnable from the devcontainer, so contributors can validate their changes locally before pushing.

**Changes:**

- **`.devcontainer/devcontainer.json`** — Adds the `docker-in-docker` devcontainer feature, required to run the MegaLinter Docker image from inside the devcontainer. Also migrates settings to the current `customizations.vscode` schema and adds the `ms-azuretools.vscode-docker` extension.
- **`.devcontainer/devcontainer-lock.json`** — New lock file pinning the `docker-in-docker` feature to `v2.17.0` for reproducible builds.
- **`.devcontainer/Dockerfile`** — Updates the base image to `mcr.microsoft.com/devcontainers/python:1-${VARIANT}` and removes a stale Yarn apt repository that caused `apt-get update` failures.
- **`.devcontainer/run-megalinter.sh`** — New helper script that mirrors the MegaLinter CI step locally. Running `bash .devcontainer/run-megalinter.sh` scans only changed files (matching `VALIDATE_ALL_CODEBASE: false` in CI); `VALIDATE_ALL_CODEBASE=true bash .devcontainer/run-megalinter.sh` scans everything.
- **`requirements-docs.txt`** — Pins `Pygments>=2,<2.20` to avoid an `AttributeError` introduced in Pygments 2.20.0, which caused the mkdocs build to crash when `pymdownx.highlight` passes `filename=None` to `HtmlFormatter`.
- **`lychee.toml`** — Accepts HTTP 403 and 429 responses in addition to 2xx, preventing false-positive link-check failures on sites (including GitHub) that block automated crawlers. Also removes an invalid `header` table entry and excludes additional bot-blocking domains.

## Checklist

- [x] Changes follow the repo structure and land in the appropriate folder and section
- [x] No confidential information
- [x] No duplicated content
- [ ] Labeled appropriately
- [ ] This PR was reviewed by at least one subject matter expert
- [x] No lint check errors related to your changes

> Note: You may see link check errors on pages you have not touched. This is normal, and due to either broken links or sites that reject link checker bots. The reviewer will help you get to a green state on these.


lychee.toml
```diff
@@ -49,7 +49,11 @@ retry_wait_time = 2
 # accept = "200..=204, 429"
 # accept = ["200", "429"]
 # accept = "200, 429"
-accept = ["200"]
+#
+# 403 (Forbidden) and 429 (Too Many Requests) are accepted because they indicate
+# the target page exists but the host blocks or throttles automated clients in
+# CI. Treating them as failures produces false positives for live links.
+accept = ["200..=299", "403", "429"]
```
shiranr, line 56: Globally accepting 403 is risky — 403 often indicates a genuinely removed/protected resource, not just bot-blocking. Since the PR already adds per-URL excludes for the known bot-blocked domains, prefer keeping accept narrow (e.g. ["200..=299", "429"]) and continuing to handle exceptions via exclude. Otherwise real breakage on any 403-returning host will silently pass CI.

lychee.toml
```diff
@@ -61,7 +65,7 @@ require_https = false
 method = "get"
 
 # Custom request headers
-header = ["name=value", "other=value"]
+# header = { name = "value", other = "value" }
```
shiranr, line 68: Do we really need this if it's commented out?

.devcontainer/run-megalinter.sh
```diff
@@ -0,0 +1,21 @@
...
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+VALIDATE_ALL="${VALIDATE_ALL_CODEBASE:-false}"
+
+echo "Running MegaLinter (VALIDATE_ALL_CODEBASE=${VALIDATE_ALL})..."
+docker run --rm \
+  -v "${REPO_ROOT}:/tmp/lint" \
+  -w /tmp/lint \
+  -e VALIDATE_ALL_CODEBASE="${VALIDATE_ALL}" \
+  -e DEFAULT_BRANCH=main \
+  oxsecurity/megalinter:v8.1.0
```
shiranr, line 21: v8.1.0 is hardcoded here but .github/workflows/mega-linter.yml has its own pinned version. If they drift, the local run stops mirroring CI, defeating the purpose. Consider sourcing the version from a shared file, or at least adding a comment reminding maintainers to bump both together.

.devcontainer/run-megalinter.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+# Usage:
+#   .devcontainer/run-megalinter.sh              # lint only changed files
+#   VALIDATE_ALL_CODEBASE=true .devcontainer/run-megalinter.sh  # lint everything
+#
+# Reports are written to ./megalinter-reports/ in the repo root.
+set -euo pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+VALIDATE_ALL="${VALIDATE_ALL_CODEBASE:-false}"
```
shiranr, line 13: MegaLinter's VALIDATE_ALL_CODEBASE=false mode typically needs DEFAULT_WORKSPACE=/tmp/lint set (and a mounted .git) to diff against DEFAULT_BRANCH=main. Worth testing that bash .devcontainer/run-megalinter.sh (no env override) actually detects only changed files rather than silently linting everything.

.devcontainer/run-megalinter.sh
```diff
@@ -0,0 +1,21 @@
...
+# .github/workflows/mega-linter.yml and config in .mega-linter.yml.
+#
+# Usage:
+#   .devcontainer/run-megalinter.sh              # lint only changed files
+#   VALIDATE_ALL_CODEBASE=true .devcontainer/run-megalinter.sh  # lint everything
+#
+# Reports are written to ./megalinter-reports/ in the repo root.
+set -euo pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
```
shiranr, line 12: git rev-parse --show-toplevel will error out if the script is run outside a git worktree — fine for the documented use case, but consider a friendlier error message.

.devcontainer/devcontainer.json
```diff
@@ -60,12 +68,12 @@
 	// },
 
 	// Use 'postCreateCommand' to run commands after the container is created.
-	"postCreateCommand": "pip3 install -r requirements-docs.txt",
+	"postCreateCommand": "pip3 install -r requirements-docs.txt && chmod +x .devcontainer/run-megalinter.sh",
```
shiranr, line 71: Redundant chmod +x in postCreateCommand — the file is already committed with mode 100755. The chmod is harmless but unnecessary.

requirements-docs.txt
```diff
@@ -1,6 +1,7 @@
 mkdocs==1.3.*
 mkdocs-material==8.2.*
 pymdown-extensions>=10.0
+Pygments>=2,<2.20
```
shiranr, line 4: Pygments>=2,<2.20 — reasonable stopgap. The root cause is that mkdocs==1.3.* and mkdocs-material==8.2.* are years old (current is 1.6+/9.x); this cap is a symptom. Fine to unblock CI, but worth filing a follow-up issue to modernize the docs toolchain.

requirements-docs.txt
```diff
@@ -1,6 +1,7 @@
 mkdocs==1.3.*
```
shiranr, line 1: I think we should consider updating the dependencies in this file first in a separate PR

.devcontainer/Dockerfile
```diff
@@ -1,8 +1,11 @@
...
-# See here for image contents: https://github.com/microsoft/vscode-dev-containers/blob/v0.202.3/containers/python-3/.devcontainer/base.Dockerfile
-# [Choice] Python version (use -bullseye variants on local arm64/Apple Silicon): 3, 3.9, 3.8, 3.7, 3.6, 3-bullseye, 3.9-bullseye, 3.8-bullseye, 3.7-bullseye, 3.6-bullseye, 3-buster, 3.9-buster, 3.8-buster, 3.7-buster, 3.6-buster
+# See here for image contents: https://github.com/devcontainers/images/tree/main/src/python
+# [Choice] Python version (use -bullseye variants on local arm64/Apple Silicon): 3, 3.9, 3.8, 3.7, 3.6, 3-bullseye, 3.9-bullseye, 3.8-bullseye, 3.7-bullseye, 3.6-bullseye
 ARG VARIANT=3-bullseye
-FROM mcr.microsoft.com/vscode/devcontainers/python:0-${VARIANT}
+FROM mcr.microsoft.com/devcontainers/python:1-${VARIANT}
+
+# Remove stale Yarn apt repository that has an expired GPG key and causes apt-get update failures
+RUN rm -f /etc/apt/sources.list.d/yarn.list /etc/apt/sources.list.d/yarn.list.save
```
shiranr, line 8: Yarn cleanup runs rm -f before any apt-get update. If another expired repo shows up later, contributors will hit the same class of failure. Consider running apt-get update in the same layer to fail fast at build time rather than at postCreateCommand.

